PR #5, opened 2024-05-30
🗞️ Add Safe config for EBI & Rarible

janjakubnanista:
### In this PR

- Adding contract networks config for EBI & Rarible
- Small cleanup: remove `ethers` dependency and replace with `@etgherproject/units` dev dependency
- Fix Safe URL for Rarible


calebbanister (approved):
lgtm

PR #27, opened 2024-06-13
🗞️ Add IOTA, TAIKO, SEI & XCHAIN

janjakubnanista:
### In this PR

- Add 4 new chains along with their configurations
- Add safe configs for these chains
- Add Stargate & Nethermind DVNs
- Add @LayerZero-Labs executors
- Use the fee data for SEI as it has the same gas issue as arbitrum sepolia


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -249,6 +274,9 @@ export const ASSETS: Record<TokenName, AssetConfig> = {
             [EndpointId.EBI_V2_MAINNET]: {
                 type: StargateType.Oft,
             },
+            [EndpointId.SEI_V2_MAINNET]: {
```
stgLamps: We're not deploying OFT.USDT to Sei

Should this be Taiko instead?

TRileySchwarz (approved):

calebbanister (approved):

PR #50, opened 2024-06-26
🪲 configurator queue capacity must be a divisor of 2e16

ryandgoulding:


packages/stg-evm-v2/devtools/src/token-messaging/config.ts
```diff
@@ -68,6 +73,10 @@ export const initializeBusQueueStorage: TokenMessagingConfigurator = withAsyncLo
                     .map((edge) => edge.vector.to.eid)
 
                 const queueCapacity = await sdk.getQueueCapacity()
+                assert(
+                    isDivisorOf2e16(queueCapacity),
+                    `Queue capacity must be a divisor of 2e16 (65536), got: ${queueCapacity}`
+                )
```
janjakubnanista, line 79: Well in this case it's already been configured so what's done is done, I would maybe print out a warning but not bring down this script

janjakubnanista (approved):

PR #71, opened 2024-06-29
chore:  test arbitrum mainnet deployment

ryandgoulding:


ryandgoulding:
we will not commit an intermediate version and wait for the audited one.

PR #73, opened 2024-07-02
🪲 estimateSendFee{,v2,Epv2}(...) do not take into account OFT-specific bps

ryandgoulding:
This is actually broken in the following three functions:
* estimateSendFee(...)
* estimateSendFeeV2(...)
* estimateSendFeeEpv2(...)

The latter two functions are released and deployed, so we must protect the interface.  It doesn't actually end up making a difference as the messaging fee is the same regardless of the bps configuration in the default implementation.  That said, it is not future proof for Adapter implementations in which the MessagingFee may vary based on the configured bps.

This PR fixes the implementation of all three.

### DIFF
--- a/packages/stg-evm-v2/src/peripheral/oft-wrapper/OFTWrapper.sol
+++ b/packages/stg-evm-v2/src/peripheral/oft-wrapper/OFTWrapper.sol
@@ -2,8 +2,6 @@
 
 pragma solidity ^0.8.0;
 
-import { console } from "forge-std/Console.sol";
-
 import { ReentrancyGuard } from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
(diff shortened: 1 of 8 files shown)


packages/stg-evm-v2/src/peripheral/oft-wrapper/OFTWrapper.sol
```diff
@@ -400,11 +400,12 @@ contract OFTWrapper is IOFTWrapper, Ownable, ReentrancyGuard {
 
     function estimateSendFeeEpv2(
         address _oft,
+        address _token,
         SendParamEpv2 calldata _sendParam,
         bool _payInLzToken,
         FeeObj calldata _feeObj
     ) external view returns (MessagingFeeEpv2 memory) {
-        (uint256 amount, , ) = getAmountAndFees(_oft, _sendParam.amountLD, _feeObj.callerBps);
+        (uint256 amount, , ) = getAmountAndFees(_token, _sendParam.amountLD, _feeObj.callerBps);
```
TRileySchwarz, line 408: You can just query the token from the oft, cast it as IOFT?

TRileySchwarz (approved):

PR #84, opened 2024-07-05
🧹  clean up setCallerBpsCap(...)

ryandgoulding:


packages/stg-evm-v2/src/peripheral/oft-wrapper/OFTWrapper.sol
```diff
@@ -42,7 +42,7 @@ contract OFTWrapper is IOFTWrapper, Ownable, ReentrancyGuard {
     }
 
     function setCallerBpsCap(uint256 _callerBpsCap) external onlyOwner {
-        require(_callerBpsCap < BPS_DENOMINATOR || _callerBpsCap == MAX_UINT, "OFTWrapper: callerBpsCap >= 100%");
+        require(_callerBpsCap <= BPS_DENOMINATOR, "OFTWrapper: callerBpsCap > 100%");
```
ravinagill15, line 45: Just wondering if we should add this validation to the constructor as well?

TRileySchwarz (approved):

packages/stg-evm-v2/src/peripheral/oft-wrapper/OFTWrapper.sol
ryandgoulding: Great point!  Ideally we would.  In this case, Stargate DAO is the only (only one I know of) instantiator of this contract, and we are familiar with the initial constructor arguments and how they should be used.  If we were adding this as a standard for other folks, then absolutely, we should likely be a bit more careful to protect our users from making a bad setting.

In this case I preferred an external method for `setCallerBpsCap(...)`, which can't be called from the constructor.  I did this because in general, `external` calls end up being cheaper than `public` ones.  In a "normal" programming language, we would just call the setter from the constructor.  In this case though, given the circumstances, I think we are fine.

That said, in general you are absolutely right.  I like your thinking on this.

ravinagill15: Understood, thanks for the explanation!

PR #114, opened 2024-08-08
🗞️ Update the default base bus ride gas limit

TRileySchwarz:


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -131,7 +131,7 @@ export const DEFAULT_TOKEN_MESSAGING_NETWORK_CONFIG: TokenMessagingNetworkConfig
     nativeDropAmount: BigInt(1e13),
     taxiGasLimit: 150000n,
     busGasLimit: 38000n,
-    busRideGasLimit: 30000n,
+    busRideGasLimit: 43000n,
```
ryandgoulding, line 134: Why is this creeping up so much?

ryandgoulding (approved):

packages/stg-definitions-v2/src/constant.ts
TRileySchwarz: turns out it was very under charged in many places, and people keep on sending the big bus rides causing failures. Its an issue of the upper limit being caught multiple times~

janjakubnanista (approved):

calebbanister (commented):
ok

PR #125, opened 2024-08-26
update allocations for September

stgLamps:
Updated rewarder.config with new allocations for the stakingTokens for V2 incentives (for September 2024)

### DIFF
--- a/.changeset/silly-coats-compete.md
+++ b/.changeset/silly-coats-compete.md
@@ -1,5 +1,0 @@
----
-"@stargatefinance/stg-evm-v2": major
----
-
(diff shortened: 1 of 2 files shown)


.changeset/silly-coats-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stargatefinance/stg-evm-v2": major
```
TRileySchwarz, line 2: No changeset is neccessary, and specifically the 'major' flag is improperly used

TRileySchwarz (approved):

PR #131, opened 2024-10-08
🧹 Specify confirmations for EBI

janjakubnanista:
### In this PR

- Add `confirmations` to configuration types for token & credit messaging
- Set the confirmations to `50` for EBI


calebbanister (approved):
increase EBI block confs, lgtm

PR #132, opened 2024-10-09
🗞️ Coredao mainnet configs

stgLamps:
In this PR:

- Updating deployment configs to enable USDC & USDT Pools on Coredao


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -687,6 +699,16 @@ export const REWARDS: RewardsConfig = {
             },
         },
     },
+    [RewardTokenName.CORE]: {
+        name: 'CORE',
+        networks: {
+            //
+            // Mainnet
+            [EndpointId.COREDAO_V2_MAINNET]: {
+                address: '0x0000000000000000000000000000000000000000',
```
janjakubnanista, line 708: This needs a value right?

janjakubnanista (changes requested):

packages/stg-definitions-v2/src/constant.ts
stgLamps: CORE = native gas token

calebbanister (approved):
i have updated the things that needed updating and reviewed it all. lgtm

janjakubnanista (approved):

PR #140, opened 2024-10-17
🗞️ Degen mainnet config

stgLamps:
**In this PR**

- Degen chain mainnet config


janjakubnanista (approved):
Since we are adding USDC in this PR, #135 needs to be merged _OR_ a more nuanced approach to wiring must be taken. The #135 outlines the approach in code so I would prefer to merge it to avoid any mistakes or complications.

USDC wiring is easier when the USDC is configured with a hot wallet, then the ownership is transferred, then the rest of the configuration follows. This avoids having to sign many more transactions as a multisig

janjakubnanista:
Please resolve the conflicts

ravinagill15 (approved):

PR #147, opened 2024-10-23
🧹 Integrating USDT deployment into stargate

ravinagill15:
## In this PR
- Referencing deployed USDT proxy address in `stg-definitions-v2/constant.ts`


packages/stg-definitions-v2/src/utils.ts
```diff
@@ -24,3 +24,18 @@ export const getNetworkConfig = (endpointId: EndpointId): NetworkConfig => {
...
+export const getUSDTAddress = (usdtData: any): string => {
+    try {
+        if (usdtData.proxies && usdtData.proxies.length > 0) {
+            const lastProxy = usdtData.proxies[usdtData.proxies.length - 1]
+            return lastProxy.address ?? ''
+        } else {
+            console.log(`No proxies found in ${JSON.stringify(usdtData)}`)
+        }
+    } catch (error) {
+        console.error(`Error loading data: ${error}`)
```
ravinagill15, line 37: I will use logger in an upcoming commit instead

PR #148, opened 2024-10-23
🧹 Disable gas profiling workflows because we don't like them anymore

janjakubnanista:
### In this PR

- Before we investigate the performance degradation of the gas profiling workflows, we disable them


calebbanister (approved):
turning off for now because it takes too long to build packages and we can manually do profiling if we need to

PR #164, opened 2024-10-31
🧹 Update readme.md

defitricks:
In the description of the `parseError` function: "This function parses a given string for error and decodes it." It should be: "This function parses a given string for **an** error and decodes it."


janjakubnanista (approved):

defitricks:
[janjakubnanista](https://github.com/janjakubnanista) merge it, pls.

PR #171, opened 2024-11-05
🧹 Add contract and function name to Asset logs

ravinagill15:
## In this PR:

- Addressing [STAR2-1015](https://layerzerolabs.atlassian.net/browse/STAR2-1015)
- Passing metadata to OmniTransaction within sdks to improve logging when printed

image


packages/stg-devtools-evm-hardhat-v2/src/asset/sdk.ts
```diff
@@ -45,9 +45,22 @@ export class Asset extends Ownable implements IAsset {
...
+        //         Add function being called to log
+        //              likely as an optional param to OmniTransaction first
+        //              then add support to logger?
+
+        // where/how is omnitransaction logged?
+
         return {
             ...this.createTransaction(data),
             description: `Setting address config for ${formatOmniPoint(omniContractToPoint(this.contract))}:\n${printRecord(config)}`,
+            // TODO functionName: 'setAddressConfig',
```
ravinagill15, line 63: I am investigating the best way to add this additional logging.

1. For contract name, I see that `OmniTransaction.OmniPoint` has a `contractName` property that is already fetched with the `formatOmniPoint` method in `devtools` [here](https://github.com/LayerZero-Labs/devtools/blob/36446d0eba516bed98748ea49009efc7b6dafec4/packages/devtools/src/omnigraph/format.ts#L6-L7). Perhaps I can re-use this function within the printed table to replace `OmniAddress`?
2. For function name, I am thinking to add an optional property to `OmniTransaction` so we can return it alongside description [here](https://github.com/stargate-protocol/stargate-v2/blob/26feafc1659f82e0428373e29220e22a7ba5b1ce/packages/stg-devtools-evm-hardhat-v2/src/asset/sdk.ts#L78). Then it should automatically be included in the printed table I think.

My main blocker right now is that I am unsure of which logic _prints_ the below table. It seems to take `OmniTransaction` and prints it in a pretty way. I need to make the following changes to that table logic:
1. Instead of printing `OmniAddress` it should print the contract name (perhaps using `formatOmniPoint`)
2. Also prints `functionName`, which will be an additional property of `OmniTransaction`

image

janjakubnanista: The SDK should already have a `label` property that you can use instead of just the `OmniPoint`. 

Besides that, the `OmniPoint` can already have `label` or `contractName` I believe or smth like that on it (if not we can add it). The idea for the SDKs in general is that they are not tied to `hardhat` so just keep that in mind - they can reference a contract without having its source code.

ravinagill15: Hmm I think I see what you mean -- I am able to access `this.point` in this file, but the `contractName` property is `undefined` (`address` is defined). 

I am looking into where `this.point` is coming from to see if anything needs to be changed there. Calling `omniContractToPoint(this.contract).contractName` also returns `undefined`. 

In the meantime, I have opened up a PR in devtools repo that updates the properties of `OmniTransaction` so that the above log can display all of its properties if you have time to review 🙏  https://github.com/LayerZero-Labs/devtools/pull/991

I will close this PR out soon because I think all changes can be done in devtools -- I believe @calebbanister would like these logs for _every_ transaction so updating a single description in a single sdk config is likely not the way to go i think

ravinagill15: Ok so this PR will not be closed actually because we need to pass `functionName`. I am still determining why `point.contractName` is undefined!

packages/stg-devtools-evm-hardhat-v2/src/blacklistable/mixin.ts
```diff
@@ -44,6 +46,8 @@ export const BlacklistableMixin: IBlacklistable = {
         return {
             ...this.createTransaction(data),
             description: `Setting ${address} blacklist status to ${blacklisted}`,
+            functionName: 'setBlacklisted',
+            functionArgs: `address = ${address}\nblacklisted = ${blacklisted}`, // TODO confirm this pattern with Jan before proceeding to make this change in all files
```
ravinagill15, line 50: @janjakubnanista I have a devtools PR up [here](https://github.com/LayerZero-Labs/devtools/pull/991/files) that adds these 2 properties. While deploying stargate, caleb mentioned it would be best if we could always display:

1. contract name
  - I believe this can be accessed through `Omnipoint.contractName` but that is currently undefined so looking into it
2. function name and functionArgs
  - As you can see above i added both of these as a new property on OmniTransaction in the devtools PR. If you confirm this pattern is acceptable, I will do this for all sdks where transactions are created. I see that `functionArgs` clashes with `description` in some cases, so I will try to remove redundancy as much as possible when updating the other functions.

packages/stg-devtools-evm-hardhat-v2/src/asset/sdk.ts
```diff
@@ -47,7 +47,8 @@ export class Asset extends Ownable implements IAsset {
 
         return {
             ...this.createTransaction(data),
-            description: `Setting address config for ${formatOmniPoint(omniContractToPoint(this.contract))}:\n${printRecord(config)}`,
+            description: `Setting address config for ${formatOmniPoint(omniContractToPoint(this.contract))}`,
+            metadata: { functionName: 'setAddressConfig', functionArgs: `config = ${printRecord(config)}` },
```
ravinagill15, line 51: This is currently throwing an error because I need to update the devtools dependency once the [devtools PR ](https://github.com/LayerZero-Labs/devtools/pull/991/files)is merged and published

packages/stg-devtools-evm-hardhat-v2/src/asset/sdk.ts
```diff
@@ -47,7 +55,12 @@ export class Asset extends Ownable implements IAsset {
 
         return {
             ...this.createTransaction(data),
-            description: `Setting address config for ${formatOmniPoint(omniContractToPoint(this.contract))}:\n${printRecord(config)}`,
+            description: `Setting address config for ${formatOmniPoint(omniContractToPoint(this.contract))}`,
+            metadata: {
+                contractName: this.contractName,
+                functionName: 'setAddressConfig',
+                functionArgs: `config = ${printRecord(config)}`,
+            },
```
ravinagill15, line 63: Currently not printing the new fields, even after cleaning cache, installing dependencies, and building. Will continue investigating!

Screenshot 2024-11-19 at 3 50 04 PM

ravinagill15: But appears in devtools when wiring:

Screenshot 2024-11-20 at 10 59 20 AM

ravinagill15: waitamin

network is the difference here...perhaps I need to update the formatting elsewhere

ravinagill15: ok i think stg actually uses the [devtools-evm-hardhat version of fomatOmniTransaction](https://github.com/LayerZero-Labs/devtools/blob/447af656cd2e6fdfb4890fc3a4995d2089fd045a/packages/devtools-evm-hardhat/src/transactions/format.ts#L8)

so I likely need to update devtools-evm-hardhat's _devtools_ dependency to be the latest

ravinagill15: The reason the wire task showed the updated logs is because it imports `formatOmniTransaction` from `devtools` instead of from `devtools-evm-hardhat`

ravinagill15: hmm wait but we set the devtools dependency for devtools-evm-hardhat as `0.4.0` so it should still pick up `0.4.1` since it is just a patch change...

ravinagill15: Hmm updating `devtools-evm-hardhat` to have `devtools` version with the updated logs did not seem to fix this. The new metadata fields don't seem to print in the stargate repo but they print in the devtools repo. 

I made sure to delete all artifacts/cache/any other build files and ran `pnpm i && pnpm build` before trying to log an omnitransaction:
Screenshot 2024-11-21 at 11 50 05 AM

PR #179, opened 2024-11-09
new proxy usdc for peaq

stgLamps:
In this PR

- New USDC Proxy USDC address for peaq


ravinagill15:
Peaq deployed with correct USDT address [here](https://github.com/stargate-protocol/stargate-v2/pull/176)

PR #181, opened 2024-11-13
🧹 Add transferTokenOwnership function to StargateOFT

ravinagill15:
The Tether USDT contracts require the owner of the token contract to be the only minter allowed. For stargate, this means that the `StargateOFTUSDT` contract will be the owner and minter of the USDT token. 

However the `StargateOFTUSDT` contract currently does not support all `onlyOwner` functions of the USDT token contract -- it only supports mint and burn. Therefore, we will implement a `transferTokenOwnership` function to allow the option of temporarily transferring ownership if the unsupported `onlyOwner` functions on the USDT token contract are ever needed.

This is the extent of the change -- we do not need to worry about the eventual transfer of ownership to Tether which will revoke `StargateOFTUSDT`'s minting privileges because at that point, `StargateOFTUSDT` will be deprecated anyway in favour of `StargatePoolUSDT`.

### DIFF
--- a/packages/stg-evm-v2/src/StargateOFT.sol
+++ b/packages/stg-evm-v2/src/StargateOFT.sol
@@ -22,6 +22,12 @@ contract StargateOFT is StargateBase {
         address _owner
     ) StargateBase(_token, IERC20Metadata(_token).decimals(), _sharedDecimals, _endpoint, _owner) {}
 
+    /// @notice Transfer ownership of the token to a new owner.
+    /// @param _newOwner The account to set as owner
+    function transferTokenOwnership(address _newOwner) external onlyOwner {
+        IERC20Minter(token).transferOwnership(_newOwner); // TODO how do i know this is the correct function to call -- where exactly does this lead after the IERC20Minter?
+    }
+
     /// @notice Burn tokens to represent their removal from the local chain
     /// @param _from The address to burn tokens from
     /// @param _amount How many tokens to burn in LD
--- a/packages/stg-evm-v2/src/StargateOFT.sol
+++ b/packages/stg-evm-v2/src/StargateOFT.sol
@@ -12,10 +12,6 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 /// @title A Stargate contract representing an OFT. This contract will burn OFTs when sending tokens
 /// @title to other chains and mint tokens when receiving them from other chains.
 contract StargateOFT is StargateBase {
-    // error TransferTokenOwnershipFailed();
-
-    // event TokenOwnershipTransferred(address token, address newOwner);
-
     /// @notice Create a StargateOFT contract administering an OFT.
     /// @param _token The OFT to administer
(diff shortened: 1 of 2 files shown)


packages/stg-evm-v2/src/StargateOFT.sol
```diff
@@ -22,6 +22,12 @@ contract StargateOFT is StargateBase {
         address _owner
     ) StargateBase(_token, IERC20Metadata(_token).decimals(), _sharedDecimals, _endpoint, _owner) {}
 
+    /// @notice Transfer ownership of the token to a new owner.
+    /// @param _newOwner The account to set as owner
+    function transferTokenOwnership(address _newOwner) external onlyOwner {
+        IERC20Minter(token).transferOwnership(_newOwner); // TODO how do i know this is the correct function to call -- where exactly does this lead after the IERC20Minter?
+    }
+
```
ravinagill15, line 30: I have 2 questions here:
1. See todo -- how do i know this is the correct function to call? Where exactly does this lead after the IERC20Minter? I feel like I would need to update another contract like [ERC20Token](https://github.com/stargate-protocol/stargate-v2/blob/main/packages/stg-evm-v2/src/mocks/ERC20Token.sol) (this one is just for testing tho), [OFTTokenERC20](https://github.com/stargate-protocol/stargate-v2/blob/main/packages/stg-evm-v2/src/utils/OFTTokenERC20.sol), or [OFTTokenERC20Upgradeable](https://github.com/stargate-protocol/stargate-v2/blob/main/packages/stg-evm-v2/src/utils/OFTTokenERC20Upgradeable.sol).
2. Is it okay to update `StargateOFT.sol` or should I create a new `StargateOFTUSDT.sol` that extends `StargateOFT.sol` and adds this `transferTokenOwnership` function? I updated `StargateOFT.sol` for now thinking it may avoid downstream changes to deploy/config scripts, but please let me know if this is not recommended! I made it `onlyOwner` for added security as well

TRileySchwarz: 1) You would simply test that it works, and then you know its fine
2) Im leaning towards just implementing it for all tokens, and it will just fail for non usdt ones

janjakubnanista: Also you can use Ownable(token) instead of IErc20Minter

packages/stg-evm-v2/src/interfaces/IERC20Minter.sol
janjakubnanista, line 1: I would rather use Ownable when casting the token, @TRileySchwarz ?

packages/stg-evm-v2/src/StargateOFT.sol
ravinagill15: Amazing, thank you both! Just tested this out after deploying USDT and stargate on testnet (and configuring) and it works like a charm!

janjakubnanista (approved):

PR #182, opened 2024-11-14
🧹 Updating external USDT deployment process

ravinagill15:
## In this PR:
- Added config step to update token ownership of USDT to `StargateOFTUSDT` so that it can be the minter (only owner can mint)
- Added example of usdt external deployment format
- Added peaq USDT config script that will guide external USDT configs moving forward

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,9 @@ configure-testnet:
 	# Transfer ownership
 	$(TRANSFER_OWNERSHIP) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/usdc-token.config.ts --signer deployer
 
+	# Transfer USDT ownership
+	$(TRANSFER_OWNERSHIP) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/usdt-token.config.ts --signer deployer
+
 	# Configure the assets
 	$(CONFIGURE_ASSET) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/asset.usdc.config.ts --signer deployer
 	$(CONFIGURE_ASSET) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/asset.usdt.config.ts --signer deployer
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -185,6 +185,9 @@ configure-testnet:
 	# Transfer ownership
 	$(TRANSFER_OWNERSHIP) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/usdc-token.config.ts --signer deployer
 
+	# Transfer USDT ownership
+	$(TRANSFER_OWNERSHIP) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/usdt-token.config.ts --signer deployer
```
ravinagill15, line 189: I am wondering if anything else needs to change in the makefile regarding USDT. For example,  in oft-token.config.ts I see [USDT-related configs for minters](https://github.com/stargate-protocol/stargate-v2/blob/46c7642a6e4aeff0cd1d6bfb4959dfd61918895b/packages/stg-evm-v2/devtools/config/mainnet/01/oft-token.config.ts#L168-L175) -- so should these now be ignored for external deployments or are the USDT configs in this file still needed?

packages/stg-evm-v2/devtools/config/mainnet/01/usdt-token.config.ts
```diff
@@ -0,0 +1,46 @@
...
+import { onPeaq } from '../utils'
+
+const fiatContract = { contractName: 'FiatTokenV2_2' }
+
+// For external USDT deployments
+const usdtPeaqAsset = getAssetNetworkConfig(EndpointId.PEAQ_V2_MAINNET, TokenName.USDT)
+assert(usdtPeaqAsset.address != null, `External USDT address not found for PEAQ`)
+
+// TODO change USDCNodeConfig to USDTNodeConfig, but that does not yet exist
+export default async (): Promise<OmniGraphHardhat<USDCNodeConfig, unknown>> => {
```
ravinagill15, line 19: Is it worth adding a `USDTNodeConfig` if this script works? I think we should but unsure of how big of a lift it is. Will look into this

Makefile
janjakubnanista: If we don't need those configs then we can remove them and the same goes for any other tokens - if they come preconfigured then we're good

packages/stg-evm-v2/devtools/config/mainnet/01/usdt-token.config.ts
janjakubnanista: I think there should be an `OwnableConfig` type already for this in `ua-devtools`

janjakubnanista (changes requested):

Makefile
ravinagill15: Do you mean remove them as in, not even add them to the configuration files when deploying stargate (i.e [the PRs](https://github.com/stargate-protocol/stargate-v2/pull/140/files) @stgLamps has been putting up)?

ravinagill15: Considering this will only be for chains moving forward now ^

For chains with existing deployments, like klaytn, we can remove the configs when we redeploy perhaps

janjakubnanista: For any and all tokens that come preconfigured

janjakubnanista (approved):

PR #129, opened 2024-09-24
🧹 Change compiler version for USDC contracts

ravinagill15:
## In this PR:
Moving forward, we need to deploy the USDC contracts with the version that uses compiler `v0.6.12+commit.27d51765`. This is necessary because only this specific version allows us to transfer ownership to the usdc admin and therefore this is blocking us from launching on new chains.

The circle engineers have pointed us to [this FiatTokenProxy](https://arbiscan.io/token/0xaf88d065e77c8cC2239327C5EDb3A432268e5831#code) contract on arb as the source of truth, which corresponds to the [FiatTokenV2_2 implementation contract here](https://arbiscan.io/address/0x86e721b43d4ecfa71119dd38c0f938a75fdb57b3#code), which references this [SignatureChecker library contract here](https://arbiscan.io/address/0x4e7D093EE4d74a01905Cf5CA92eB0bf154a53247#code)

There are a couple ways to change the USDC contracts:
1. Manually copy and paste the older versions of the contracts to reference in deploy scripts

      - Very tedious as there are many misaligned dependencies

3. [implemented in current PR] In the deploy script, the raw bytecode of the older versions of the contracts will be deployed

PR #184, opened 2024-11-18
🧹 Change compiler version for USDC contracts (Reverts #129)

ravinagill15:
This reverts commit [65c78b5ffefc683f5a6167a1166760d5f4aa0cbf](https://github.com/stargate-protocol/stargate-v2/pull/129). 

The reverted commit allowed us to deploy USDC with just the raw bytecode, which is not an ideal approach. Since then, we have decided on a better approach -- we will always deploy USDC from [circle's repo](https://github.com/circlefin/stablecoin-evm) and configure the address of the USDC Proxy within stargate [(see here)](https://github.com/stargate-protocol/stargate-v2/pull/176/files#diff-ba4479457e246e8302ee62e180e60373999e1673da83a5715b046d0ef8d45092R443)


packages/stg-evm-v2/hardhat.config.ts
```diff
@@ -118,7 +118,7 @@ const networks: NetworksUserConfig = {
     //
     'arbsep-testnet': {
         eid: EndpointId.ARBSEP_V2_TESTNET,
-        url: process.env.RPC_URL_ARBITRUM_TESTNET || 'https://arbitrum-sepolia.gateway.tenderly.co',
+        url: process.env.RPC_URL_ARBITRUM_TESTNET || 'https://arbitrum-sepolia.blockpi.network/v1/rpc/public',
```
janjakubnanista, line 121: Are these changes intentional?

janjakubnanista (approved):

packages/stg-evm-v2/hardhat.config.ts
ravinagill15: Yeah these were just done to make the testnet build pass -- the previous RPCs were causing it to fail

PR #185, opened 2024-11-18
🧹 Add metadata to Blacklistable logs

ravinagill15:
## In this PR:

- Addressing [STAR2-1015](https://layerzerolabs.atlassian.net/browse/STAR2-1015)

### DIFF
--- a/packages/stg-devtools-evm-hardhat-v2/src/blacklistable/mixin.ts
+++ b/packages/stg-devtools-evm-hardhat-v2/src/blacklistable/mixin.ts
@@ -22,6 +22,11 @@ export const BlacklistableMixin: IBlacklistable = {
         return {
             ...this.createTransaction(data),
             description: `Setting blacklister to address ${blacklister}`,
+            metadata: {
+                contractName: this.contractName,
+                functionName: 'setBlacklister',
+                functionArgs: `blacklister = ${blacklister}`,
+            },
         }
     },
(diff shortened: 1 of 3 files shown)


packages/stg-devtools-evm-hardhat-v2/src/blacklistable/mixin.ts
```diff
@@ -22,6 +22,11 @@ export const BlacklistableMixin: IBlacklistable = {
         return {
             ...this.createTransaction(data),
             description: `Setting blacklister to address ${blacklister}`,
+            metadata: {
+                contractName: this.contractName,
```
ravinagill15, line 26: Not recognized here, but recognized in sdk... looking into this

ravinagill15: maybe can pass as param

ravinagill15: or change parameter `this` to be `Blacklistable` instead of `OmniSDK`, but that may break the call from usdc to blacklistable [here](https://github.com/stargate-protocol/stargate-v2/blob/bc089aca096c8ca0cb167fce0b3f1225b431f68f/packages/stg-devtools-evm-hardhat-v2/src/usdc/sdk.ts#L52)

PR #200, opened 2024-11-29
🧹 Mantle Sepolia Testnet Config

ravinagill15:
## In this PR:
- Config and deployment for Mantle Sepolia Testnet based on:

```
Mantle Sepolia testnet (40246)
Pool WETH
Pool USDC
Pool USDT
```


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -664,6 +676,9 @@ export const REWARDS: RewardsConfig = {
             [EndpointId.BSC_V2_TESTNET]: {},
             [EndpointId.OPTSEP_V2_TESTNET]: {},
             [EndpointId.ARBSEP_V2_TESTNET]: {},
+            [EndpointId.MANTLESEP_V2_TESTNET]: {
+                /* TODO this is a pool*/
+            },
```
ravinagill15, line 681: I am assuming this needs a config here since we are adding a testnet with pools for the first time, but unsure of what the config should look like. Looks like we need to deploy the RewardTokenMock to Mantle Sepolia Testnet first and likely list its address here?

cc: @TRileySchwarz

packages/stg-definitions-v2/src/constant.ts
```diff
@@ -709,6 +724,13 @@ export const REWARDS: RewardsConfig = {
             [EndpointId.ZKCONSENSYS_V2_MAINNET]: {
                 address: '0x808d7c71ad2ba3FA531b068a2417C63106BC0949',
             },
+
+            //
+            // Testnet
+            //
+            [EndpointId.MANTLESEP_V2_TESTNET]: {
+                address: 'TODO?',
```
ravinagill15, line 732: Hmm this may not be needed here actually because we will be testing with RewardTokenMock

packages/stg-definitions-v2/src/constant.ts
stgLamps: This config is correct, but the TODO isn't needed. We don't actually need to deploy the mock reward token as we won't be using the famrs. 

The line above "            // We don't deploy the token on testnet since there are no pools", is meant to say "since there are no farms"

ravinagill15: Ahh i see thank you for clarifying!

packages/stg-definitions-v2/src/constant.ts
ravinagill15: Thanks!

stgLamps (approved):

PR #203, opened 2024-12-02
Bump Peaq, Degen, Rootstock Native Drop 

stgLamps:
**In this PR**

- Native Drop on Peaq was previously set to 0.01, which was too low. This PR bumps it 10x to 0.1PEAQ


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -1884,7 +1884,7 @@ export const NETWORKS: NetworksConfig = {
             ...DEFAULT_TOKEN_MESSAGING_NETWORK_CONFIG,
             requiredDVNs: [DVNS.NETHERMIND[EndpointId.ROOTSTOCK_V2_MAINNET], DVNS.STG[EndpointId.ROOTSTOCK_V2_MAINNET]],
             executor: EXECUTORS.LZ_LABS[EndpointId.ROOTSTOCK_V2_MAINNET],
-            nativeDropAmount: parseEther('0.00000000005').toBigInt(),
+            nativeDropAmount: parseEther('0.000055').toBigInt(),
```
ravinagill15, line 1887: Is bumping rootstock intentional as well?

ravinagill15 (commented):
Looks good! Let's add changeset as well

packages/stg-definitions-v2/src/constant.ts
stgLamps: yes it is - renaming PR!

PR #208, opened 2024-12-04
🧹 Add dynamic num retries

ravinagill15:
## In this PR:
- Allowing dynamic number of retries

### DIFF
--- a/packages/stg-evm-v2/hardhat.config.ts
+++ b/packages/stg-evm-v2/hardhat.config.ts
@@ -7,6 +7,7 @@ configDotenv({
     path: ['.env.local', '.env'],
 })
 
+import { AsyncRetriable } from '@layerzerolabs/devtools'
 import { EndpointId } from '@layerzerolabs/lz-definitions'
 
 import { getSafeConfig } from './devtools/config/utils'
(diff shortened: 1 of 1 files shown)


packages/stg-evm-v2/hardhat.config.ts
```diff
@@ -32,6 +33,14 @@ import '@typechain/hardhat'
 // Devtools-specific tasks
 import './devtools/tasks'
 
+AsyncRetriable.config = {
+    enabled: true,
+    numAttempts: process.env.NUM_RETRIES ? parseInt(process.env.NUM_RETRIES) : 5,
```
ItsAdel, line 38: in `@layerzerolabs/devtools`, it defaults the `numAttempts` to 3. Just noting it here and wanted to see if we want to keep in the same or update it as it is.

```
interface RetriableConfig
{
...
/**
     * Number of times the method call will be retried. The default is 3
     *
     * @default 3
     */
    numAttempts: number;
...
}
```

ItsAdel (approved):
small comment to consider but lgtm

packages/stg-evm-v2/hardhat.config.ts
ravinagill15: Good point! For stargate we have run into rpc issues where we wanted to retry even like 20 times, so I think upping it from 3 to 5 is fine for this PR given the history. In the future this will likely be overwritten by the environment variable anyway

stgLamps (approved):

PR #237, opened 2024-12-11
Abstract Testnet Deployment

ravinagill15:
## In this PR:
- Deployed and configured stargate for abstract testnet


ravinagill15:
Deployment is done, configuration is a work in progress (currently running into a gas limit issue when configuring credit messaging from abstract to klaytn)

PR #259, opened 2025-01-16
Superposition USDT Config

stgLamps:
In this PR
* Add support for bridging stgUSDT to Superposition


clauBv23:
closing, since no longer valid

PR #263, opened 2025-01-21
Soneium mainnet config

stgLamps:
In this PR
* Deploy ETH Pool 
* Deploy USDC.e Pool


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -1769,6 +1783,24 @@ export const NETWORKS: NetworksConfig = {
...
+        tokenMessaging: {
+            ...DEFAULT_TOKEN_MESSAGING_NETWORK_CONFIG,
+            requiredDVNs: [DVNS.NETHERMIND[EndpointId.SONEIUM_V2_MAINNET], DVNS.STG[EndpointId.SONEIUM_V2_MAINNET]],
+            executor: EXECUTORS.LZ_LABS[EndpointId.SONEIUM_V2_MAINNET],
+            nativeDropAmount: parseEther('0.0005').toBigInt(),
+        },
+        safeConfig: {
+            safeAddress: '0xc38C19FC8E2e66715dd80b2b8192A30e36dD68af',
+            //TODO : verify what the API is for this Superchain Safe? Below is not correct
+            safeUrl: 'https://safe.optimism.io/',
```
ravinagill15, line 1801: Seems like https://safe.optimism.io/welcome/accounts?chain=minato is for their testnet? Got this from their docs [here](https://docs.soneium.org/docs/users/wallets#safe) but nothing for mainnet is listed.

Also, do we need to add any `contractNetworks` config like [this one](https://github.com/stargate-protocol/stargate-v2/blob/c74d364bf3595273a3b277f9e6813473eb27589c/packages/stg-definitions-v2/src/constant.ts#L1124-L1134) (I don't think so since we are working with pools, but want to confirm)

ravinagill15 (commented):
Looks good! Just left a minor comment, will approve once we determine the safeURL

packages/stg-definitions-v2/src/constant.ts
stgLamps: updated

ravinagill15 (approved):

PR #270, opened 2025-01-23
Bera mainnet config

stgLamps:
In this PR
* Config USDC.e to Bera
* Config WETH to Bera


packages/stg-definitions-v2/src/constant.ts
```diff
@@ -1100,6 +1113,23 @@ export const NETWORKS: NetworksConfig = {
...
+        tokenMessaging: {
+            ...DEFAULT_TOKEN_MESSAGING_NETWORK_CONFIG,
+            requiredDVNs: [DVNS.NETHERMIND[EndpointId.BERA_V2_MAINNET], DVNS.STG[EndpointId.BERA_V2_MAINNET]],
+            executor: EXECUTORS.LZ_LABS[EndpointId.BERA_V2_MAINNET],
+            nativeDropAmount: parseEther('0.001').toBigInt(),
+        },
+        safeConfig: {
+            safeAddress: '0x5A185C81AFa796a753dd662343cbF911254f796e',
+            safeUrl: 'https://transaction.bp.w3us.site/',
+        },
```
ravinagill15, line 1131: Do we need to add a `contractNetworks` object here similar to codex?

ravinagill15 (approved):
LGTM, one minor question

packages/stg-definitions-v2/src/constant.ts
stgLamps: we do not - it's an official safe deployment

PR #291, opened 2025-02-11
Ebi Chain Shutdown

stgLamps:
Ebi Chain shutdown , there are certain configs that must be updated to facilitate this. 

* Setting DVNs to 0x000


packages/stg-evm-v2/devtools/config/mainnet/utils.ts
```diff
@@ -13,7 +13,7 @@ export const onBsc = withEid(EndpointId.BSC_V2_MAINNET)
 export const onCodex = withEid(EndpointId.CODEX_V2_MAINNET)
 export const onCoredao = withEid(EndpointId.COREDAO_V2_MAINNET)
 export const onDegen = withEid(EndpointId.DEGEN_V2_MAINNET)
-export const onEbi = withEid(EndpointId.EBI_V2_MAINNET)
+// export const onEbi = withEid(EndpointId.EBI_V2_MAINNET)
```
ravinagill15, line 16: Since this is used in other config files, we should also comment it out in those files

calebbanister (approved):

PR #322, opened 2025-02-27
refactor assets config (testnet and mainnet)

clauBv23:
This PR refactors the asset configuration for both mainnet and testnet.

The new approach dynamically defines contracts, points, and nodes based on valid chains, rather than specifying them individually.

A valid chain is determined by the intersection of all defined chains (acting as the source of truth, stored in [the utils file](https://github.com/stargate-protocol/stargate-v2/blob/3251089e89388bcc9f8f63b66f8ddf65b477caa1/packages/stg-evm-v2/devtools/config/mainnet/utils.ts#L129)) and the chains that should be excluded from specific configurations. This simplifies the process by only requiring exceptions to be specified. However, this logic can be adjusted with minimal effort if needed.

All return values for contracts and connections are now auto-generated. When adding a new chain, it only needs to be included in the source of truth list with the corresponding EIP and added to the exception list if it should be excluded from an asset configuration.


ravinagill15 (approved):
🚀

ravinagill15 (approved):

PR #323, opened 2025-02-28
ci: refactor feelib v1 config

clauBv23:
following #322 idea, this PR is doing the same for the feelib v1 config files


ravinagill15 (approved):
🚀

PR #324, opened 2025-03-03
feat: refactor the oft wrappers in mainnet

clauBv23:
following the same idea that in #323 and #322, this PR refactors oft wrapper config file. 

While doing this refactor, I noticed that `coredao-mainnet` was not being configured in [oft-wrapper.config](https://github.com/stargate-protocol/stargate-v2/blob/main/packages/stg-evm-v2/devtools/config/mainnet/01/oft-wrapper.config.ts), this refactor also includes it.


ravinagill15 (approved):
It's beautiful 🥹

PR #328, opened 2025-03-11
Add Abstract & Bera RPC

stgLamps:
Both RPCs are no longer private, can add public RPCs into main


clauBv23:
closed in favor of #329

PR #329, opened 2025-03-11
feat: change some broken rpc urls and add validation in makefile

clauBv23:


ravinagill15 (commented):
🚀 

I think we should wait to merge these changes until the devtools package is updated in npm at which point we should update the dependency in `package.json` and run `pnpm i` to update the lock file as well

.github/workflows/reusable-configure.yaml
```diff
@@ -20,12 +20,12 @@ env:
   # We'll add all the RPC URLs from secrets
   #
   # Mainnet
-  RPC_URL_ABSTRACT_MAINNET: ${{ secrets.RPC_URL_ABSTRACT_MAINNET }}
+  # RPC_URL_ABSTRACT_MAINNET: ${{ secrets.RPC_URL_ABSTRACT_MAINNET }}
```
ravinagill15, line 23: Should we just remove these from the file instead of commenting them out?

packages/stg-evm-v2/.env
```diff
@@ -1,5 +1,5 @@
 # Used in fork tests for StargateZapperV1
-RPC_URL_ETHEREUM_MAINNET=https://rpc.ankr.com/eth
+RPC_URL_ETHEREUM_MAINNET=https://eth.llamarpc.com
```
ravinagill15, line 2: llama rpcs should generally be avoided (I forget the issue but something with it not being fully compatible with our deployment process)

packages/stg-evm-v2/hardhat.config.ts
```diff
@@ -300,7 +300,7 @@ const networks: NetworksUserConfig = {
     },
     'ethereum-mainnet': {
         eid: EndpointId.ETHEREUM_V2_MAINNET,
-        url: process.env.RPC_URL_ETHEREUM_MAINNET || 'https://rpc.ankr.com/eth',
+        url: process.env.RPC_URL_ETHEREUM_MAINNET || 'https://eth.llamarpc.com',
```
ravinagill15, line 303: Same not about llama rpcs

packages/stg-evm-v2/hardhat.config.ts
```diff
@@ -495,7 +495,7 @@ const networks: NetworksUserConfig = {
     },
     'polygon-mainnet': {
         eid: EndpointId.POLYGON_V2_MAINNET,
-        url: process.env.RPC_URL_POLYGON_MAINNET || 'https://polygon.meowrpc.com',
+        url: process.env.RPC_URL_POLYGON_MAINNET || 'wss://polygon-bor-rpc.publicnode.com',
```
ravinagill15, line 498: does websocket rpc work as expected?

packages/stg-evm-v2/hardhat.config.ts
```diff
@@ -731,6 +731,7 @@ const hardhatNetworks: Pick<HardhatUserConfig, 'networks'> = {
             accounts: sandboxAccounts,
             blockGasLimit: 30_000_000,
             throwOnCallFailures: false,
+            allowUnlimitedContractSize: true,
```
ravinagill15, line 734: How come we added this? Is it to remove some kind of warning?

ravinagill15 (commented):
Look great! Just a couple minor comments

.github/workflows/reusable-configure.yaml
clauBv23: yeah, removed https://github.com/stargate-protocol/stargate-v2/pull/329/commits/5bbdc28b7d9686643ec1b9048f1c15cce2535828

packages/stg-evm-v2/.env
clauBv23: ankr one was not working that's why I changed it.

Updating to a no llama one https://github.com/stargate-protocol/stargate-v2/pull/329/commits/751c0a3c7b9050f28ced2e9dd2fcd80b2a0c8d8b

packages/stg-evm-v2/hardhat.config.ts
clauBv23: it has worked perfectly on my side but, changing it to ensure there is no weird behaviour

https://github.com/stargate-protocol/stargate-v2/pull/329/commits/71f2751472a50307600a4f2f158f0bed7a9a9e93

packages/stg-evm-v2/hardhat.config.ts
ravinagill15: Ah interesting -- I don't see a direct change that would impact contract size, but this flag is only really for development on local nodes (sandboxes) anyway. Mainnet / testnet evms have a 24kb limit and we don't deploy endpointv2mock there anyway so I think this is fine

ravinagill15 (approved):

packages/stg-evm-v2/hardhat.config.ts
clauBv23: yeap, good point. will inspect further this to flag the exact issue

PR #331, opened 2025-03-13
Avalanche Fuji testnet config

clauBv23:
### In this PR:

Config for Avalanche Fuji Testnet (43113) based on: 

- Pool USDT

todo:
- ~~Confirm that the USDT address on the chain is correct.~~  => will deploy new one during deployment


ravinagill15 (approved):
Looks good! Let's rebase before merging

PR #341, opened 2025-03-19
Avalanche fuji deployment

clauBv23:


packages/stg-devtools-v2/src/messaging/config.ts
```diff
@@ -15,7 +15,7 @@ import { IMessaging, MessagingOmniGraph } from './types'
 
 export type MessagingConfigurator = Configurator<MessagingOmniGraph, IMessaging>
 
-const createMessagingLogger = () => createModuleLogger('TokenMessaging')
+const createMessagingLogger = () => createModuleLogger('CreditMessaging')
```
ravinagill15, line 18: How come this was changed?

packages/stg-evm-v2/hardhat.config.ts
ravinagill15, line 1: I think we need to add avalanche config to hardhat config

packages/stg-devtools-v2/src/messaging/config.ts
clauBv23: I realized that this logger tag was wrong since it was being used by both token and credit messaging 

Check [here](https://github.com/stargate-protocol/stargate-v2/blob/eba19c44c914983c960bdeb3766a2f12e411fa57/packages/stg-devtools-v2/src/token-messaging/config.ts#L15)  and [here](https://github.com/stargate-protocol/stargate-v2/blob/eba19c44c914983c960bdeb3766a2f12e411fa57/packages/stg-devtools-v2/src/credit-messaging/config.ts#L3) 

this was causing the logs to show token messaging tag on the credit messaging configuration

It is not correct either tag it as credit messaging, so I tagged it simply as message https://github.com/stargate-protocol/stargate-v2/pull/341/commits/56d0113f45810e890662854dd0247c1afc8f15e5

packages/stg-evm-v2/hardhat.config.ts
clauBv23: it was already done on the config pr 
https://github.com/stargate-protocol/stargate-v2/blob/eba19c44c914983c960bdeb3766a2f12e411fa57/packages/stg-evm-v2/hardhat.config.ts#L134-L138

ravinagill15: Ah nice!

ravinagill15 (approved):

PR #345, opened 2025-03-26
Plume phoenix mainnet config

clauBv23:


packages/stg-evm-v2/package.json
```diff
@@ -106,7 +106,7 @@
     "@layerzerolabs/protocol-devtools-evm": "~1.2.1",
     "@layerzerolabs/solidity-examples": "^1.1.0",
     "@layerzerolabs/test-devtools": "~0.2.9",
-    "@layerzerolabs/toolbox-foundry": "~0.1.7",
```
clauBv23, line 109: changing this because after the `0.1.8` version the file structure changes from `solidity-bytes-utils/contracts/BytesLib.sol` to `solidity-bytes-utils/BytesLib.sol`, since the change was marked as patch and version was defined with `~` pnpm was bumping `toolbox-foundry` version and was not compiling

PR #350, opened 2025-04-01
allow to pass --continue flag to script

clauBv23:


Makefile
```diff
@@ -46,6 +46,12 @@ DEPLOY_ARGS=
 # These allow consumers of this script to pass flags like --ci
 CONFIGURE_ARGS_COMMON=
 
+# Arguments to be always passed to the validate RPCs task
+# 
+# These allow consumers of this script to pass flags like --continue
+VALIDATE_RPCS_ARGS=
```
ravinagill15, line 52: Will this work as expected considering we did not update the task in devtools to accept the `--continue` flag yet?

clauBv23: > we did not update the task in devtools

What you mean with this? 

I added the flag in the task, check [here](https://github.com/clauBv23/devtools/blob/d5664adcf1245c90ad1a43bb89f36423fe88d5dc/packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts#L196)

ravinagill15 (approved):

PR #351, opened 2025-04-02
Manta mainnet config

clauBv23:
### In this PR:

Config for Manta Pacific Mainnet (169) based on: 

- Native pool

TODOs: 
- [x] Add native drop amount
- [x] Add rewards and staking


packages/stg-definitions-v2/src/constant.ts
ravinagill15, line 1: Do we need to add oft wrapper config?

ravinagill15 (approved):
🚀

PR #359, opened 2025-04-10
Xdc mainnet deployment

clauBv23:


packages/stg-evm-v2/devtools/config/mainnet/01/shared.ts
ravinagill15, line 1: Were the removals in this file intentional?

ravinagill15 (approved):

packages/stg-evm-v2/devtools/config/mainnet/01/shared.ts
clauBv23: yes by mistake I added the LP token and all this deployment is all ofts

PR #360, opened 2025-04-14
Nibiru mainnet config

clauBv23:
### In this PR:

Config for Nibiru (6900) based on: 

- USDC.e - oft
- USDT - oft
- WETH - oft 

TODOs: 
- [x] Add native drop amount
- [x] Deploy USDC and add the address
- [x] Deploy USDT and add the address
- [x] Add safe multisig address
- [x] Add safe deployment


stgLamps (approved):
lgtm!

PR #366, opened 2025-04-24
Feat: add wiring checks

clauBv23:
In this PR:
- added hardhat tasks for checking that the wiring is correct on the current mesh
- added workflow action to check this in the pipeline 
- updated the Harhdat chain configs to get the lz RPC aggregator URL


ravinagill15 (approved):
🚀

PR #386, opened 2025-07-18
📤 [deploy] Sophon mainnet

clauBv23:


.changeset/silent-swans-pay.md
ravinagill15, line 1: Let's add all packages here so that the planner can access the deployments within the `stg-evm-sdk-v2` package

ravinagill15 (commented):
Looks good, let's update changeset to include all packages so we can avoid having followup PR

Let's hold off on merging until configuration is done

.changeset/silent-swans-pay.md
clauBv23: done https://github.com/stargate-protocol/stargate-v2/pull/386/commits/e5313de8a13c836d391f6593750fa11eb74f910f

ravinagill15 (approved):

PR #411, opened 2025-08-20
🛑 [shutdown] XChain mainnet

clauBv23:


packages/stg-definitions-v2/src/constant.ts
ravinagill15, line 1: Once the configuration has been updated in the next run, we should make a follow up PR to remove the discontinued chains from this file (including past chains like ebi, plume (legacy), etc.

ravinagill15 (approved):

packages/stg-evm-v2/hardhat.config.ts
```diff
@@ -626,13 +626,6 @@ const networks: NetworksUserConfig = {
         safeConfig: getSafeConfig(EndpointId.UNICHAIN_V2_MAINNET),
         timeout: DEFAULT_NETWORK_TIMEOUT,
     },
-    'xchain-mainnet': {
-        eid: EndpointId.XCHAIN_V2_MAINNET,
-        url: process.env.RPC_URL_XCHAIN_MAINNET || 'https://xchain-rpc.idex.io/',
-        accounts: mainnetAccounts,
-        safeConfig: getSafeConfig(EndpointId.XCHAIN_V2_MAINNET),
-        timeout: DEFAULT_NETWORK_TIMEOUT,
-    },
```
ravinagill15, line 635: Actually just noticed -- I think we may need this to set the dvn/executor for xchain to the zero address

PR #417, opened 2025-08-25
🛑 [shutdown] Odyssey testnet

clauBv23:


packages/stg-definitions-v2/src/constant.ts
ravinagill15, line 1: There is [one more reference to odyssey testnet](https://github.com/stargate-protocol/stargate-v2/blob/738dbfda366440f5c544badb3206b8ca75d8d43a/packages/stg-definitions-v2/src/constant.ts#L2965-L2968) in this file we should remove

clauBv23: but, the NetworksConfig [on mainnet](https://github.com/stargate-protocol/stargate-v2/blob/124b2a4f75be8a724c849e47a75f3989b845e70d/packages/stg-definitions-v2/src/constant.ts#L1523-L1551) has not been removed in this PR, Should it be?

PR #429, opened 2025-09-04
Bump somnia gas limit

clauBv23:


stgLamps (dismissed):
changeset but otherwise lgtm

PR #441, opened 2025-09-30
Add off chai checker GH action

clauBv23:


kmk142789 (approved):

.github/workflows/offchain-checker.yaml
```diff
@@ -0,0 +1,26 @@
+name: Offchain Checker
+
+on:
+  workflow_dispatch:
+
+jobs:
+  validate:
+    name: Run stg-evm-sdk-v2 validation
+    runs-on: ubuntu-latest
```
ravinagill15: I think we should use the `ubuntu-latest-4xlarge` runner because this validation is a heavier task with tasks being run in parallel

PR #462, opened 2025-12-10
Revert "Dvn migration from STG to LZ_Labs (#457)"

clauBv23:
## Context 
DVN migration is blocked due to not being able to propose to OneSig. Given the tight deadline, we’ll wire GateLayer first and resume the migration afterward.

## In this PR 

- Reverted #458.
- @ensdomains/buffer version updated to `0.1.3`


stgLamps (approved):
nice

PR #523, opened 2026-02-25
Sei paths unwiring 

clauBv23:
## In this PR

- Define Ethereum as only allowed peer for Sei. 

 
As a side task, we needed to unwire all Sei paths except Ethereum <-> Sei. For this, we had to create a `messaging.unwire.yml` with this info

```
rules:
  - chain: sei-mainnet
    allowed_peers:
      - ethereum-mainnet
```

But it is not included in this PR to avoid adding a new PR to remove it, since it should only be temporary.

The bundle to unwire is already proposed and should be signed in the following hours


stgLamps (approved):

ravinagill15 (approved):

clauBv23:
closing in favor of https://github.com/stargate-protocol/stargate-v2/pull/566

PR #532, opened 2026-04-01
feat: add CreditMessagingMintableBurnable for emergency credit recovery

clauBv23:
## Problem

Credits in Stargate V2 act as a capacity gate — they must be held on a chain before tokens can be redeemed or sent cross-chain. When a chain is deprecated without properly returning its credits to origin chains, those credits become permanently lost. This leaves pools with real token balances but depleted path credits, blocking all user exit paths.

**Example:** Sei USDT pool holds $27.7k but only $4.2k in usable credits — $23.5k is effectively locked with no way out.

Credits are not a security primitive (they do not custody funds — every actual fund movement requires independent authorization via LP tokens or a valid cross-chain token message). They are a liveness primitive.

## Solution

`CreditMessagingMintableBurnable` extends `CreditMessaging` with two owner-only admin functions. Both operate **locally only** — no LayerZero message is ever sent, no cross-chain coordination is required, and no LZ fees are incurred.

**`mintCredits(CreditBatch[], reason)`** — increases credits on the current chain by calling `ICreditMessagingHandler.receiveCredits(0, credits)` directly, bypassing the normal credit validation. Used to restore lost credits on affected chains.

**`burnCredits(TargetCreditBatch[], reason)`** — decreases credits on the current chain by calling `ICreditMessagingHandler.sendCredits(0, targets)` locally, stopping before `_lzSend`. Uses `minAmount = amount` (all-or-nothing), reverting if the path has insufficient credits. Used to correct over-minted credits.

Both functions require a non-empty plain-text `reason` string that is emitted on-chain as a permanent audit trail, as per the team discussion.

## Safety

- Credits cannot move funds on their own. Over-injecting local credits beyond pool balance causes failed transactions — the pool physically cannot send what it does not hold. Over-burning locks funds in the pool (same situation as the bug being fixed).
- `mintCredits` and `burnCredits` are `onlyOwner` (multisig), not planner-accessible.
- `sendCredits` remains planner-accessible and unchanged.
- Treasury fees and `recoverToken` are unaffected — they operate on `poolBalanceSD` and `treasuryFee`, which credits never touch.
- No LZ message means no DVN dependency, no peer configuration needed, and no cross-chain reorg risk.

## Changes

- **New contract** — `src/messaging/CreditMessagingMintableBurnable.sol`
Extends `CreditMessaging` with `mintCredits` and `burnCredits`.

- **New interface** — `src/interfaces/ICreditMessagingMintableBurnable.sol`
Exposes the admin functions, the `CreditMessagingMintableBurnable_EmptyReason` error, and the `CreditsMinted` / `CreditsBurned` events.

- **Deploy script** — `deploy/003-deploy-message.ts`
Non-alt chains now deploy `CreditMessagingMintableBurnable` under the existing `CreditMessaging` deployment name. Alt chains continue deploying `CreditMessagingAlt` unchanged.

- **Unit tests** — `test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol`
Extends `CreditMessagingTest` to prove backward compatibility — all parent tests run against `CreditMessagingMintableBurnable`. CreditMessagingMintableBurnable-specific coverage includes: access control, empty reason revert, unavailable asset revert, local-only operation (no `PacketSent` event), multiple batches with multiple credits per batch, partial burns (handler returns less than requested), and empty credits arrays.

- **Integration tests ** — test/integration/messaging/Messaging.CreditMintableBurnable.Integration.t.sol
Wired to real StargatePool and StargateOFT instances. Verifies that paths[eid].credit actually changes after mintCredits/burnCredits and that the LZ endpoint emits no events, confirming all operations are strictly local.

- **Existing credit messaging tests** — `test/unitest/messaging/Messaging.Credit.t.sol`
Added tests asserting that `mintCredits` and `burnCredits` produce an empty revert (function does not exist) on the base `CreditMessaging` contract.

## TBD
- ~Should we create a similar contract for alt chains? For now `CreditMessagingAlt` is unchanged — the mesh can be balanced using a single `CreditMessagingMintableBurnable` deployment~ No we will deploy on a single chain that will not be an alt chain.
- There are [invariant tests](https://github.com/stargate-protocol/stargate-v2/blob/1d37fcc3a4ab719930487df3a5a78376142510cb/packages/stg-evm-v2/test/invariant/StargateInvariantTest.t.sol#L103-L117) that check that total credits equal pool balance.
Mint/burn intentionally breaks this invariant — but so did the original bug: disconnecting a chain without returning credits already left the mesh in a state where total credits no longer matched pool balances. We decided against adjusting the invariant tests since these are emergency-only functions and the fuzzer is better spent on the normal credit flow. Gathering team thoughts on this.


packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
```diff
@@ -257,6 +257,29 @@ contract CreditMessagingTest is Test {
...
+
+    function test_MintCredits_NotAvailableOnCreditMessaging() public {
+        (bool success, ) = address(messaging).call(
+            abi.encodeWithSignature(
+                "mintCredits(uint32,(uint16,(uint32,uint64)[])[]memory,string)",
+                0,
+                new bytes(0),
+                ""
+            )
+        );
```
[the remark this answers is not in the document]
clauBv23: fixed e4f4b23951a46d04a4bc3b0c918b335f92fc5c05

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
```diff
@@ -257,6 +257,29 @@ contract CreditMessagingTest is Test {
...
+        assertFalse(success);
+    }
+
+    function test_BurnCredits_NotAvailableOnCreditMessaging() public {
+        (bool success, ) = address(messaging).call(
+            abi.encodeWithSignature("burnCredits((uint16,(uint32,uint64,uint64)[])[],string)", new bytes(0), "")
+        );
```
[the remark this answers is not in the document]
clauBv23: fixed e4f4b23951a46d04a4bc3b0c918b335f92fc5c05

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,60 @@
...
+            TargetCreditBatch calldata targetBatch = _creditBatches[i];
+            Credit[] memory burned = ICreditMessagingHandler(_safeGetStargateImpl(targetBatch.assetId)).sendCredits(
+                0,
+                targetBatch.credits
+            );
+            emit CreditsBurned(targetBatch.assetId, burned, _reason);
+        }
+    }
+
+    function quoteMintCredits(
```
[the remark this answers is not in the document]
clauBv23: check added 1ccb363d616abc8997a61334940a613d4e0122fc

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,60 @@
...
+        CreditBatch[] calldata _batches,
+        string calldata _reason
+    ) external payable onlyOwner {
+        (bytes memory message, bytes memory options) = _buildMessagePayload(_dstEid, _batches);
+        _lzSend(_dstEid, message, options, MessagingFee(msg.value, 0), msg.sender);
+        emit CreditsMinted(_dstEid, _batches, _reason);
+    }
+
+    /// @dev Credits are deducted on the current chain without sending an LZ message to any other chain.
+    function burnCredits(TargetCreditBatch[] calldata _creditBatches, string calldata _reason) external onlyOwner {
```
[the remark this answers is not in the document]
clauBv23: check added 1ccb363d616abc8997a61334940a613d4e0122fc

packages/stg-evm-v2/src/interfaces/IMintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,40 @@
...
+    /// @notice Emitted when credits are burned locally without sending an LZ message.
+    event CreditsBurned(uint16 assetId, Credit[] credits, string reason);
+
+    /// @notice Mints credits on a destination chain without consuming local credits.
+    /// @param _dstEid The destination LayerZero endpoint ID.
+    /// @param _batches The credit batches to mint. Set srcEid = dstEid to restore local redemption credits.
+    /// @param _reason Plain-text explanation emitted on-chain for auditability.
+    function mintCredits(uint32 _dstEid, CreditBatch[] memory _batches, string calldata _reason) external payable;
+
+    /// @notice Quotes the LZ messaging fee for a mintCredits call.
```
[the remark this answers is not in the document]
clauBv23: changed ab7ab809ad38f677d2c06cff3c60cf4162fdfcc6

packages/stg-evm-v2/src/interfaces/ICreditMessagingRecovery.sol
```diff
@@ -0,0 +1,40 @@
...
+/// @title MintBurnCreditMessaging API
+/// @dev Defines the emergency credit management functions available only to the contract owner.
+///      These functions are intentionally separate from the standard ICreditMessaging interface
+///      to make clear that they bypass normal credit validation and must be used with care.
+interface IMintBurnCreditMessaging {
+    /// @notice Emitted when credits are minted to a destination chain without local deduction.
+    event CreditsMinted(uint32 dstEid, CreditBatch[] batches, string reason);
+
+    /// @notice Emitted when credits are burned locally without sending an LZ message.
+    event CreditsBurned(uint16 assetId, Credit[] credits, string reason);
```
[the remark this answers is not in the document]
clauBv23: the interface includes the events https://github.com/stargate-protocol/stargate-v2/blob/f664aedbb11b0a891c428b9022b5872b30239ba0/packages/stg-evm-v2/src/interfaces/IMintBurnCreditMessaging.sol#L15-L19

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,64 @@
...
+    function mintCredits(
+        uint32 _dstEid,
+        CreditBatch[] calldata _batches,
+        string calldata _reason
+    ) external payable onlyOwner {
+        if (bytes(_reason).length == 0) revert MintBurnCreditMessaging_EmptyReason();
+        (bytes memory message, bytes memory options) = _buildMessagePayload(_dstEid, _batches);
+        _lzSend(_dstEid, message, options, MessagingFee(msg.value, 0), msg.sender);
+        emit CreditsMinted(_dstEid, _batches, _reason);
+    }
```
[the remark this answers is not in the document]
clauBv23: intentionally omitted. `mintCredits` and `burnCredits` are emergency owner-only operations that require careful review by signers and engineers before execution, including a mandatory non-empty reason. The inputs are expected to be validated off-chain before the multisig signs. This check for dummies seems not needed imo

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    function _mockStargateSendCredits(TargetCredit[] memory _credits) internal {
+        Credit[] memory actual = new Credit[](_credits.length);
+        for (uint256 i = 0; i < _credits.length; i++) {
+            actual[i] = Credit(_credits[i].srcEid, _credits[i].amount);
+        }
+        vm.mockCall(
+            STARGATE_IMPL,
+            abi.encodeWithSelector(ICreditMessagingHandler.sendCredits.selector),
```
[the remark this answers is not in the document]
clauBv23: solved 57b0eaf8252eba710bc98064ec7d63126855b075

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,64 @@
...
+import { MessagingFee } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oft/interfaces/IOFT.sol";
+
+import { CreditMessaging } from "./CreditMessaging.sol";
+import { TargetCreditBatch } from "../interfaces/ICreditMessaging.sol";
+import { CreditMsgCodec, CreditBatch } from "../libs/CreditMsgCodec.sol";
+import { IMintBurnCreditMessaging } from "../interfaces/IMintBurnCreditMessaging.sol";
+import { ICreditMessagingHandler, Credit } from "../interfaces/ICreditMessagingHandler.sol";
+
+contract MintBurnCreditMessaging is CreditMessaging, IMintBurnCreditMessaging {
+    error MintBurnCreditMessaging_EmptyReason();
```
ravinagill15: should this be in the interface?

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
```diff
@@ -0,0 +1,200 @@
...
+    function test_BurnCredits_EmitsEventAndDoesNotSendLzMessage(
+        uint32 _srcEid,
+        uint64 _amount,
+        uint64 _minAmount
+    ) public {
+        vm.assume(_amount > 0 && _minAmount <= _amount);
+        TargetCreditBatch[] memory batches = _buildBurnBatches(_srcEid, _amount, _minAmount);
+        _mockStargateSendCredits(0, batches[0].credits);
+
+        vm.expectEmit(false, false, false, false);
```
ravinagill15: Since this is going to occur locally we can verify the credits were burned too

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
ravinagill15, line 1: Should we add the following test cases?

1. `mintCredits` and `burnCredits` with multiple asset batches or multiple credits per batch? 
2. `tryDecreaseCredit` can return less than requested (or 0 if insufficient). Currently the test mocks the handler to always return the full amount. A test where the mock returns a reduced `Credit[]` would verify the event emits the actual burned amounts, not the requested ones 
3. `mintCredits` with empty batches array: what happens if you call `mintCredits(dstEid, [], "reason")`? It would encode an empty message and call `_lzSend`. Is that intended or should it revert? Worth either testing that it works or adding a guard. 
4. `burnCredits` with empty batches array, similar question, calling with an empty array emits nothing and does nothing. Probably fine, but perhaps can be an explicit test?

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    constructor(address _endpoint, address _owner) CreditMessaging(_endpoint, _owner) {}
+
+    // ---------------------------------- Only Owner ------------------------------------------
+
+    /// @dev Credits are sent to the destination chain without being deducted on the current chain.
+    function mintCredits(
```
ravinagill15: nit: should we make this nonreentrant?

ravinagill15 (commented):
Nice 🚀 let's add a changeset as well

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint32 _dstEid,
+        CreditBatch[] calldata _batches,
+        string calldata _reason
+    ) external payable onlyOwner {
+        if (bytes(_reason).length == 0) revert MintBurnCreditMessaging_EmptyReason();
+        (bytes memory message, bytes memory options) = _buildMessagePayload(_dstEid, _batches);
+        _lzSend(_dstEid, message, options, MessagingFee(msg.value, 0), msg.sender);
```
tinom9: I think the system becomes cleaner if we keep both operations local:

- `mintCredits` calls `receiveCredits`.
- `burnCredits` calls `sendCredits`.

That way, indexers can easily know which credits were credited after a corresponding debit on source, and which were minted without the debit.

Also, we take out of the equation LZ messaging logic and keep operations atomic, which seems like a huge feature.

Do you see any downside?

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,64 @@
...
+        emit CreditsMinted(_dstEid, _batches, _reason);
```
tinom9: Unrelated to the code, but have we thought about how the planner/indexer will handle the new event topology?

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
```diff
@@ -0,0 +1,64 @@
...
+
+import { MessagingFee } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oft/interfaces/IOFT.sol";
+
+import { CreditMessaging } from "./CreditMessaging.sol";
+import { TargetCreditBatch } from "../interfaces/ICreditMessaging.sol";
+import { CreditMsgCodec, CreditBatch } from "../libs/CreditMsgCodec.sol";
+import { IMintBurnCreditMessaging } from "../interfaces/IMintBurnCreditMessaging.sol";
+import { ICreditMessagingHandler, Credit } from "../interfaces/ICreditMessagingHandler.sol";
+
+contract MintBurnCreditMessaging is CreditMessaging, IMintBurnCreditMessaging {
```
tinom9: I'd consider renaming to `CreditMessagingRecoverable`, `CreditMessagingAdjustable`, or `CreditMessagingMintableBurnable` to keep the consistency with Stargate contract names:

1. Suffixes over prefixes. This feels like something we should respect.
2. Use-cases over actions (`StargatePoolMigratable` instead of `StargatePoolBurnable`). This is more subjective and I don't have a strong opinion.

We could also consider renaming `mintCredits` and `burnCredits` to `mintCreditsForRecovery` and `burnCreditsForRecovery`. No strong opinion here either, just sharing my thoughts.

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
```diff
@@ -257,6 +259,26 @@ contract CreditMessagingTest is Test {
         );
     }
 
+    // ---------------------------------- mintCredits / burnCredits not available ------------------------------------------
+
+    function test_MintCredits_NotAvailableOnCreditMessaging() public {
+        CreditBatch[] memory batches = new CreditBatch[](0);
+        (bool success, ) = address(messaging).call(
```
tinom9: Let's assert the revert reason has length 0 so that we know it's not any other error (authentication, validations).

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
```diff
@@ -257,6 +259,26 @@ contract CreditMessagingTest is Test {
...
+            abi.encodeWithSelector(IMintBurnCreditMessaging.mintCredits.selector, uint32(0), batches, "")
+        );
+        assertFalse(success);
+    }
+
+    function test_BurnCredits_NotAvailableOnCreditMessaging() public {
+        TargetCreditBatch[] memory batches = new TargetCreditBatch[](0);
+        (bool success, ) = address(messaging).call(
```
tinom9: Idem.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
```diff
@@ -0,0 +1,200 @@
...
+        messaging.mintCredits(DST_EID, batches, "");
+    }
+
+    function test_RevertIf_BurnCredits_EmptyReason() public {
+        TargetCreditBatch[] memory batches = _buildBurnBatchesForAsset(ASSET_ID, 100, 0);
+        vm.expectRevert(MintBurnCreditMessaging.MintBurnCreditMessaging_EmptyReason.selector);
+        messaging.burnCredits(batches, "");
+    }
+
+    function test_RevertIf_MintCreditsByNonOwner(address _nonOwner) public {
```
tinom9: Name formatting seems inconsistent `test_RevertIf_MintCreditsByNonOwner` vs `test_RevertIf_MintCredits_ByNonOwner`.

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
tinom9: :+1:

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
```diff
@@ -0,0 +1,200 @@
...
+        CreditBatch[] memory batches = _buildMintBatches(1, 100);
+        vm.expectRevert(MintBurnCreditMessaging.MintBurnCreditMessaging_EmptyReason.selector);
+        messaging.mintCredits(DST_EID, batches, "");
+    }
+
+    function test_RevertIf_BurnCredits_EmptyReason() public {
+        TargetCreditBatch[] memory batches = _buildBurnBatchesForAsset(ASSET_ID, 100, 0);
+        vm.expectRevert(MintBurnCreditMessaging.MintBurnCreditMessaging_EmptyReason.selector);
+        messaging.burnCredits(batches, "");
+    }
```
tinom9: This should go in the `burnCredits` section?

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
tinom9: Agree.

I'd also consider being more exhaustive in the event assertion.

```suggestion
        vm.expectEmit(true, true, true, true, address(messaging));
```

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
```diff
@@ -0,0 +1,200 @@
...
+import { CreditMessaging, MessagingBase } from "../../../src/messaging/CreditMessaging.sol";
+import { CreditMessagingOptions } from "../../../src/messaging/CreditMessagingOptions.sol";
+import { CreditBatch } from "../../../src/libs/CreditMsgCodec.sol";
+import { IOAppCore } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/interfaces/IOAppCore.sol";
+import { MessagingFee, MessagingReceipt } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oft/interfaces/IOFT.sol";
+import { ILayerZeroEndpointV2 } from "@layerzerolabs/lz-evm-protocol-v2/contracts/interfaces/ILayerZeroEndpointV2.sol";
+import { AddressCast } from "@layerzerolabs/lz-evm-protocol-v2/contracts/libs/AddressCast.sol";
+import { LzUtil } from "../../layerzero/LzUtil.sol";
+
+contract MintBurnCreditMessagingTest is Test {
```
tinom9: Since this contract is append-only, (doesn't change existing functionality), I'd extend `CreditMessagingTest` here to make sure all of the existing tests hold.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
tinom9: Agree.

> What happens if you call mintCredits(dstEid, [], "reason")? It would encode an empty message and call _lzSend. Is that intended or should it revert?

`sendCredits` succeeds, so I believe we shouldn't revert here.

If we were to call `_lzSend` (as per my previous messages, I don't think we should), `sendCredits` skips the `_lzSend`, so we should here too.

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
tinom9: It has the `onlyOwner` modifier, so every potential re-entrancy would need to come from the owner.

This seems completely safe.

Plus, for consistency, `sendCredits` doesn't implement re-entrancy guards.

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
```diff
@@ -0,0 +1,200 @@
...
+        vm.expectRevert(MessagingBase.Messaging_Unavailable.selector);
+        messaging.burnCredits(batches, "test");
+    }
+
+    function test_BurnCredits_EmitsEventAndDoesNotSendLzMessage(
+        uint32 _srcEid,
+        uint64 _amount,
+        uint64 _minAmount
+    ) public {
+        vm.assume(_amount > 0 && _minAmount <= _amount);
```
tinom9: We should always prefer `bound` over `vm.assume`.

`bound` reuses the inputs, whereas `vm.assume` rejects them.

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    function test_RevertIf_BurnCredits_UnavailableAsset(uint16 _assetId) public {
+        vm.assume(_assetId != ASSET_ID && _assetId > 0);
+        TargetCreditBatch[] memory batches = _buildBurnBatchesForAsset(_assetId, uint64(100), uint64(0));
+        vm.expectRevert(MessagingBase.Messaging_Unavailable.selector);
+        messaging.burnCredits(batches, "test");
+    }
+
+    function test_BurnCredits_EmitsEventAndDoesNotSendLzMessage(
```
tinom9: We're not really testing LZ message is not being sent.

We could test it.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
```diff
@@ -0,0 +1,200 @@
...
+
+    function test_SendCredits_StillAccessibleByPlanner(uint32 _srcEid, uint64 _amount) public {
+        vm.assume(_amount > 0);
+        TargetCreditBatch[] memory batches = _buildBurnBatches(_srcEid, _amount, 0);
+        _mockStargateSendCredits(DST_EID, batches[0].credits);
+        _mockEndpointSend(100);
+
+        vm.prank(PLANNER);
+        messaging.sendCredits(DST_EID, batches);
+    }
```
tinom9: Can remove if we inherit the parent's test suite.

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
clauBv23: done 358fc24b63109dfa0c7d6d43acd1ade1cbe3acb4

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
clauBv23: agree

changed 6431dbbdf8e7cbe9e92c49997fa799bb75d22103

clauBv23: opened a discussion with the team, but I think this solution is way better

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
clauBv23: you mean when new credits were created? 

I'm not sure we have an indexer here. I think the planner just queries all the information on-chain and gets the real credits in the mesh. Will double check

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
clauBv23: I agree, however, I renamed to `CreditMessagingMintableBurnable`, I agree with _"Use-cases over actions"_ but I don't like `CreditMessagingRecoverable` or `CreditMessagingAdjustable` and I cannot find a better name that describes the use case 🤔

clauBv23: Renamed to `CreditMessagingRecovery` it's a credit messaging contract that performs recovery. 
658e3804208ca52ecb4b703d14094df705b052ad

Kept the function names as `mintCredits` and `burnCredits` for now, I don't like adding the for recovery, it seems too verbose.

 I have been thinking about `restoreCredits` / `trimCredits`, first one implies bringing something back to what it should be, and second implies removing excess, both suggest the recovery intent. But not sure if it would make it harder to understand if the credits are increasing or decreasing. wdyt @tinom9 ?

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
clauBv23: done 5ec98132376a209f9b5fd9659e0914d08d5b355a

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
clauBv23: done 5ec98132376a209f9b5fd9659e0914d08d5b355a

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
clauBv23: solved 83d6a47bfc02402f649ab180abf0f50120764686

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
clauBv23: done 83d6a47bfc02402f649ab180abf0f50120764686

skipped `test_MintCredits_NotAvailableOnCreditMessaging`  and `test_BurnCredits_NotAvailableOnCreditMessaging` since both functions are defined in `CreditMessagingRecovery`

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
clauBv23: done e829f10649e020b49dab9ab6dda891deec2c2f85

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
clauBv23: removed 83d6a47bfc02402f649ab180abf0f50120764686

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
clauBv23: added 7c43a2ab15106211f85215bf0ff0af8ded093f16

It now checks the lz msg send event is not emitted

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
clauBv23: fixed 09e3f45e520be11afd8eaa8286aeb117972c096c

All fucntion names now follow the pattern `test_RevertIf_[FunctionName]_[condition]`

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
clauBv23: checking credits where `increased`/`decreased` when they were `minted`/`burned`, will mean checking the `receiveCredits` and `sendCredits` functions do what they are supposed to do, which should already be checked in the Pool tests. 

Since these are unit tests, I would keep this function to only check the correct function is called on each option, and the correct events are emitted. 

It would be good to have integration tests for this but currently, the only integration tests for `CreditMessaging` [are for cross-chain](https://github.com/stargate-protocol/stargate-v2/blob/main/packages/stg-evm-v2/test/crosschain/CrossChainTest.t.sol), It does not seem correct to me to add them there. 

I can create a new file for `CreditMessagingRecovery` integration tests if you think it is important to have them, or if you think they can be in the same file, I can also add them there.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
clauBv23: tests added0456965ca40c32f7dca832c7caddf59c83435fbd

we no longer have the issue with `_lzSend` having an empty message to send since the mint is only local.

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
tinom9: I don't know how the new events will affect any backend system that we already have, since `CreditsSent` and `CreditsReceived` are emitted.

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
tinom9: `restoreCredits` and `trimCredits` seem less verbose, what do they really mean? I prefer mint-burn.

tinom9: Let's update the PR title?

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
```diff
@@ -0,0 +1,39 @@
...
+    // ---------------------------------- Only Owner ------------------------------------------
+
+    /// @dev Credits are applied directly on the current chain by calling receiveCredits locally,
+    ///      without deducting credits from any source or sending an LZ message.
+    function mintCredits(CreditBatch[] calldata _batches, string calldata _reason) external onlyOwner {
+        if (bytes(_reason).length == 0) revert CreditMessagingRecovery_EmptyReason();
+        for (uint256 i = 0; i < _batches.length; i++) {
+            CreditBatch calldata batch = _batches[i];
+            ICreditMessagingHandler(_safeGetStargateImpl(batch.assetId)).receiveCredits(0, batch.credits);
+            emit CreditsMinted(batch.assetId, batch.credits, _reason);
```
tinom9: There's the argument for 1 top-level `CreditsMinted(_batches, _reason)` instead of N events (1 per asset).

This is not a hot path, so we don't mind about gas consumption.

`StargateBase` emits per asset, so I think I slightly prefer this approach.

Would double check if there's any system (backend) that would prefer to have this as a single emit.

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    /// @dev Credits are deducted on the current chain by calling sendCredits locally with minAmount = amount,
+    ///      making it all-or-nothing. No LZ message is sent to any other chain.
```
tinom9: `minAmount` is not a slippage-like check.

```sol
uint64 decreased = paths[c.srcEid].tryDecreaseCredit(c.amount, c.minAmount)
```

```sol
    /// @notice Decrease credit for a given path, even if only a partial amount is possible.
    /// @param _path The Path for which to decrease credit
    /// @param _amountSD The amount by which try to decrease credit
    /// @param _minKept The minimum amount of credit to keep after the decrease
    /// @return decreased The actual amount of credit decreased
    function tryDecreaseCredit(
        Path storage _path,
        uint64 _amountSD,
        uint64 _minKept
    ) internal returns (uint64 decreased) {
        uint64 currentCredit = _path.credit;
        // not allowed to try to decrease unlimited credit
        if (currentCredit == UNLIMITED_CREDIT) revert Path_UnlimitedCredit();
        if (_minKept < currentCredit) {
            unchecked {
                uint64 maxDecreased = currentCredit - _minKept;
                decreased = _amountSD > maxDecreased ? maxDecreased : _amountSD;
                _path.credit = currentCredit - decreased;
            }
        }
    }
```

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
```diff
@@ -0,0 +1,315 @@
...
+        CreditBatch[] memory batches = _buildMintBatches(1, 100);
+        vm.prank(PLANNER);
+        vm.expectRevert(OWNABLE_ERROR);
+        CreditMessagingRecovery(address(messaging)).mintCredits(batches, MINT_REASON);
+    }
+
+    function test_RevertIf_MintCredits_UnavailableAsset(uint16 _assetId) public {
+        _assetId = uint16(bound(_assetId, ASSET_ID + 1, type(uint16).max));
+        Credit[] memory credits = new Credit[](1);
+        credits[0] = Credit(1, 100);
```
tinom9: Generally prefer passing structs instead of positional parameters.

Super nit though.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
```diff
@@ -0,0 +1,315 @@
...
+        credits2[1] = Credit(4, _amount);
+        CreditBatch[] memory batches = new CreditBatch[](2);
+        batches[0] = CreditBatch(ASSET_ID, credits1);
+        batches[1] = CreditBatch(assetId2, credits2);
+
+        _mockStargateReceiveCredits(STARGATE_IMPL, credits1);
+        _mockStargateReceiveCredits(stargate2, credits2);
+
+        vm.expectCall(STARGATE_IMPL, abi.encodeCall(ICreditMessagingHandler.receiveCredits, (0, credits1)));
+        vm.expectCall(stargate2, abi.encodeCall(ICreditMessagingHandler.receiveCredits, (0, credits2)));
```
tinom9, line 103: This is consistent with the codebase, so I think we should leave as is, but I wouldn't mind the `_mock` function to also run the `vm.expectCall`.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
```diff
@@ -0,0 +1,315 @@
...
+        TargetCreditBatch[] memory batches = _buildBurnBatches(1, 100);
+        vm.expectRevert(ICreditMessagingRecovery.CreditMessagingRecovery_EmptyReason.selector);
+        CreditMessagingRecovery(address(messaging)).burnCredits(batches, "");
+    }
+
+    function test_RevertIf_BurnCredits_ByNonOwner(address _nonOwner) public {
+        vm.assume(_nonOwner != OWNER);
+        TargetCreditBatch[] memory batches = new TargetCreditBatch[](0);
+        vm.prank(_nonOwner);
+        vm.expectRevert();
```
tinom9: Revert reason?

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
```diff
@@ -0,0 +1,315 @@
...
+        CreditMessagingRecovery(address(messaging)).burnCredits(batches, BURN_REASON);
+    }
+
+    function test_RevertIf_BurnCredits_PlannerCannotBurn() public {
+        messaging.setAssetId(STARGATE_IMPL, ASSET_ID);
+        TargetCreditBatch[] memory batches = _buildBurnBatches(1, 100);
+        vm.prank(PLANNER);
+        vm.expectRevert();
```
tinom9: Idem.

packages/stg-evm-v2/src/interfaces/ICreditMessagingRecovery.sol
```diff
@@ -0,0 +1,32 @@
...
+
+    /// @notice Mints credits on the current chain by calling receiveCredits locally.
+    /// @param _batches The credit batches to mint. Set srcEid = localEid to restore local redemption credits.
+    /// @param _reason Plain-text explanation emitted on-chain for auditability.
+    function mintCredits(CreditBatch[] calldata _batches, string calldata _reason) external;
+
+    /// @notice Burns credits on the current chain by calling sendCredits locally without sending an LZ message.
+    ///         Uses minAmount = amount making it all-or-nothing — reverts if the path has insufficient credits.
+    /// @param _batches The credit batches to burn. minAmount must equal amount for each credit entry.
```
tinom9: Idem.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
```diff
@@ -0,0 +1,315 @@
...
+        batches[0] = TargetCreditBatch(ASSET_ID, credits);
+    }
+
+    function _mockStargateReceiveCredits(Credit[] memory _credits) internal {
+        _mockStargateReceiveCredits(STARGATE_IMPL, _credits);
+    }
+
+    function _mockStargateReceiveCredits(address _stargate, Credit[] memory _credits) internal {
+        vm.mockCall(_stargate, abi.encodeCall(ICreditMessagingHandler.receiveCredits, (0, _credits)), "");
+    }
```
tinom9, line 309: I don't love that we're mocking this.

It's a fundamentally important integration of this contract.

That said, this is consistent with the codebase, so I think it's okay to keep this file as is.

Can we have some integration tests though?

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
```diff
@@ -0,0 +1,315 @@
...
+        batches[0] = CreditBatch(ASSET_ID, credits);
+    }
+
+    function _buildBurnBatches(
+        uint32 _srcEid,
+        uint64 _amount
+    ) internal pure returns (TargetCreditBatch[] memory batches) {
+        batches = new TargetCreditBatch[](1);
+        TargetCredit[] memory credits = new TargetCredit[](1);
+        credits[0] = TargetCredit(_srcEid, _amount, _amount);
```
tinom9, line 299: Idem.

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
```diff
@@ -257,6 +259,28 @@ contract CreditMessagingTest is Test {
         );
     }
 
+    // ---------------------------------- mintCredits / burnCredits not available ------------------------------------------
+
+    function test_MintCredits_NotAvailableOnCreditMessaging() public virtual {
+        CreditBatch[] memory batches = new CreditBatch[](0);
+        (bool success, bytes memory data) = address(messaging).call(
+            abi.encodeWithSelector(ICreditMessagingRecovery.mintCredits.selector, uint32(0), batches, "")
```
tinom9: Is this correct?

```suggestion
            abi.encodeWithSelector(ICreditMessagingRecovery.mintCredits.selector, batches, "")
```

packages/stg-evm-v2/src/interfaces/ICreditMessagingRecovery.sol
```diff
@@ -0,0 +1,32 @@
...
+/// @title CreditMessagingRecovery API
+/// @dev Defines the emergency credit management functions available only to the contract owner.
+///      These functions are intentionally separate from the standard ICreditMessaging interface
+///      to make clear that they bypass normal credit validation and must be used with care.
+interface ICreditMessagingRecovery {
+    /// @notice Thrown when mintCredits or burnCredits is called with an empty reason string.
+    error CreditMessagingRecovery_EmptyReason();
+
+    /// @notice Emitted when credits are minted locally without a corresponding debit on a source chain.
+    event CreditsMinted(uint16 assetId, Credit[] credits, string reason);
```
tinom9: Let's index `assetId`?


```suggestion
    event CreditsMinted(uint16 indexed assetId, Credit[] credits, string reason);
```

Same for `CreditsBurned`.

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.22;
+
+import { CreditMessaging } from "./CreditMessaging.sol";
+import { TargetCreditBatch } from "../interfaces/ICreditMessaging.sol";
+import { CreditBatch } from "../libs/CreditMsgCodec.sol";
+import { ICreditMessagingRecovery } from "../interfaces/ICreditMessagingRecovery.sol";
+import { ICreditMessagingHandler, Credit } from "../interfaces/ICreditMessagingHandler.sol";
+
+contract CreditMessagingRecovery is CreditMessaging, ICreditMessagingRecovery {
```
tinom9: If we're planning to deploy this in every chain, I'd also create an alt version. If so, we don't need to use the planner to send credits, we can handle every mismatch locally.

If not, it's probably fine to keep as it.

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
clauBv23: Just confirmed the indexer does not index credits moving

packages/stg-evm-v2/src/messaging/MintBurnCreditMessaging.sol
clauBv23: updated PR  title and description. Keeping mint burn for now

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
clauBv23: this is totally a mistake on my end. It should emit a single event per function call

clauBv23: fixed 27da4963a869f16b1d2cfa29735b37a1ca66dac8 

single event emitted per function call

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
clauBv23: good catch, comment fixed f123e95d8772ccae78f20bd08141f69bb6fad98e

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
clauBv23: agree it is easier to follow, 

changed 5f13975ba435a1473cc66d6e858ceea96d96bed2

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
clauBv23: added 375a9bee14ba5ad1a8d2b8748deb8e51f8f1572c

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditRecovery.t.sol
clauBv23: added 375a9bee14ba5ad1a8d2b8748deb8e51f8f1572c

packages/stg-evm-v2/src/interfaces/ICreditMessagingRecovery.sol
clauBv23: fixed f123e95d8772ccae78f20bd08141f69bb6fad98e

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
clauBv23: added 244cf55c512dbf7fff48aea38fcc8a392d4c40e9

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
clauBv23: done 5f13975ba435a1473cc66d6e858ceea96d96bed2

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
clauBv23: the first param is `_dstEid `
https://github.com/stargate-protocol/stargate-v2/blob/d48ebfbae7a28ecf8404d4a7306abd6e1e80066c/packages/stg-evm-v2/src/interfaces/ICreditMessagingHandler.sol#L13-L15

packages/stg-evm-v2/src/interfaces/ICreditMessagingRecovery.sol
clauBv23: asset id is removed now, it would be on each CreditBatch item

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
clauBv23: we are only planning to deploy on new chains. Old ones will stay the same.

packages/stg-evm-v2/test/unitest/messaging/Messaging.CreditMintableBurnable.t.sol
clauBv23: agree we should add integratio test for this

clauBv23: added f6cd53ac842dae71c7638544338fcd0761c24707

packages/stg-evm-v2/test/integration/messaging/Messaging.CreditRecovery.Integration.t.sol
```diff
@@ -0,0 +1,142 @@
...
+    uint16 internal constant ASSET_ID = 1;
+    uint32 internal constant SRC_EID = 10;
+    string internal constant MINT_REASON = "restoring lost credits";
+    string internal constant BURN_REASON = "correcting over-minted credits";
+
+    CreditMessagingRecovery internal recoveryMessaging;
+    StargateBaseTestC internal pool;
+
+    function setUp() public {
+        pool = new StargateBaseTestC();
```
tinom9: What about testing behaviour with both `StargatePool` and `StargateOFT`, instead of `StargateBase` mocks?

packages/stg-evm-v2/test/unitest/messaging/Messaging.Credit.t.sol
clauBv23: fixed https://github.com/stargate-protocol/stargate-v2/pull/532/commits/788f8aab01c6ddd20d1c07769340323a82f0da59

packages/stg-evm-v2/test/integration/messaging/Messaging.CreditRecovery.Integration.t.sol
clauBv23: added 6b56b756824e1338e4f2c83cae5fc5514a418f69

packages/stg-evm-v2/test/unitest/messaging/Messaging.MintBurnCredit.t.sol
clauBv23: integration tests added 6b56b756824e1338e4f2c83cae5fc5514a418f69

ravinagill15 (dismissed):

tinom9 (commented):
Contracts look good to me!

packages/stg-evm-v2/src/messaging/CreditMessagingRecovery.sol
```diff
@@ -0,0 +1,41 @@
...
```
TRileySchwarz: Only comment I would make is I dont like the name CreditMessagingRecovery, really should be something like CreditMessagingV2 imo.

Further to that, I think to simplify the system and risk of deploying on new chains, that perhaps we deploy this on arbitrum only, and not new chains moving forward. We dont need to know that before sending to audit, and also regarding the contract name, not a huge deal to change mid audit as well. The rest of impl makes sense. 

Another thing to note is that currently if the planner has the contract paused, then we cant mint or burn credits. So this creates a bit of an issue if it MUST remain paused during the mint.burn recovery or migration / chain deprecation. However I think this is fine becaause worst case the owner can 1: transfer 'planner' to the owner, 2: unpause 3: mint/burncredits, 4: pause, 5: transfer planner back to the planner. This creates an atomic situation albeit, quite complicated.

tinom9: My intuition is that we only want to deploy once.

if so, I don't love `CreditMessagingV2` naming, since it implies the next version (V2) of the same thing (credit messaging). I'd stick to use-case.
- `StargatePool` -> `StargatePoolMigratable`.
- `CreditMessaging` -> `CreditMessagingRecovery` / `CreditMessagingRecoverable`.

clauBv23: I also think this should be deployed on a single chain. @tironiigor’s plan was to deploy the new version from now on, but I agree we could deploy it on one of the main chains and continue using the current one elsewhere.

Regarding the name, I don’t have a strong opinion. I’m not a big fan of CreditMessagingRecovery either, but I don’t have a better suggestion. That said, I agree with @tino that if we go with a single-chain deployment, calling it CreditMessagingV2 could be misleading since we’d deploy V1 to new chains. 

On the planner pausing point,good observation. The owner can ultimately take over the role and mint/burn if needed. It’s not ideal, but it’s possible, and everything can be done atomically. So I’d keep it in mind, but I don’t see it as an issue

TRileySchwarz: Yeah, fair if only 1 deployment shouldnt be v2, but we can brainstorm a better name. Igor agrees on doing it only one chain, as long as the migration (vs. just a new version on expansion) isnt too chaotic. Looking into that now

clauBv23: Final decision: rename to `CreditMessagingMintableBurnable`

done:
- https://github.com/stargate-protocol/stargate-v2/pull/532/commits/83063478e234ae36b594ef7c47b825dcf538dcb4 main contract and interface 
- https://github.com/stargate-protocol/stargate-v2/pull/532/commits/e38b188f5555e6322bce782a2fbecabd6178b055 tests

tinom9 (approved):

PR #538, opened 2026-04-20
fix: support comma-separated NEW_CHAIN without duplicating cross-new-chain edges

ravinagill15:
## Summary

Fixes a bug introduced alongside #514 where configuring more than one new chain at a time (e.g. `NEW_CHAIN='subtensorevm-mainnet,ault-mainnet'`) produced **58 duplicate transactions** in the OneSig bundle between the two new chains.

### What was happening

- `getNewChain()` returned the raw comma-separated string.
- `getNewChainEid()` tried to resolve the whole string as a single chain name → threw (or, in the iterate-per-chain variant, emitted each cross-new-chain edge once per run).
- `filterConnections` filtered on a single `newChainEid`, so edges between two new chains could be kept under both chains' filter passes when callers iterated manually.

### What this PR does

- `getNewChain` → `getNewChains(): string[]` (splits on `,`, trims, drops empties).
- `getNewChainEid` → `getNewChainEids(): Set<EndpointId>` — a `Set`, so edges get kept at most once.
- `filterConnections` now uses set-membership filtering. Each directed edge `A→B` is emitted exactly once if either endpoint is a new chain, whether or not both endpoints are new.
- All config builders (messaging, asset, feelib, rewarder, staking, treasurer, oft-wrapper) accept the list.
- Added multi-chain test cases to every relevant test suite.

Single-chain behavior is unchanged.

## Test plan

- [x] `pnpm test:hardhat` — 300 passing (up from 295, including 5 new multi-chain tests)
- [x] Dry-run `NEW_CHAIN='subtensorevm-mainnet,ault-mainnet' make configure-mainnet` and confirm the OneSig bundle no longer flags duplicate transactions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/stg-evm-v2/devtools/config/mainnet/01/oft-wrapper.config.ts
+++ b/packages/stg-evm-v2/devtools/config/mainnet/01/oft-wrapper.config.ts
@@ -3,7 +3,7 @@ import { OFTWrapperNodeConfig } from '@stargatefinance/stg-devtools-v2'
 import { OmniGraphHardhat } from '@layerzerolabs/devtools-evm-hardhat'
 
 import { getContractWithEid, getOneSigAddress } from '../../utils'
-import { filterValidProvidedChains, getAllChainsConfig, getNewChain, printChains } from '../../utils/utils.config'
+import { filterValidProvidedChains, getAllChainsConfig, getNewChains, printChains } from '../../utils/utils.config'
 import { setMainnetStage } from '../utils'
 
 const contract = { contractName: 'OFTWrapper' }
@@ -12,8 +12,9 @@ export default async (): Promise<OmniGraphHardhat<OFTWrapperNodeConfig, unknown>
     // Set the stage to mainnet
     setMainnetStage()
 
-    const newChain = getNewChain()
-    const chainsList = newChain ? [newChain] : process.env.CHAINS_LIST ? process.env.CHAINS_LIST.split(',') : []
+    const newChains = getNewChains()
+    const chainsList =
+        newChains.length > 0 ? newChains : process.env.CHAINS_LIST ? process.env.CHAINS_LIST.split(',') : []
 
     // get valid chains config in the chainsList
     const validChains = filterValidProvidedChains(chainsList, getAllChainsConfig())
--- a/packages/stg-evm-v2/devtools/config/mainnet/01/oft-wrapper.config.ts
+++ b/packages/stg-evm-v2/devtools/config/mainnet/01/oft-wrapper.config.ts
@@ -3,7 +3,7 @@ import { OFTWrapperNodeConfig } from '@stargatefinance/stg-devtools-v2'
 import { OmniGraphHardhat } from '@layerzerolabs/devtools-evm-hardhat'
 
 import { getContractWithEid, getOneSigAddress } from '../../utils'
-import { filterValidProvidedChains, getAllChainsConfig, getNewChains, printChains } from '../../utils/utils.config'
+import { filterValidProvidedChains, getAllChainsConfig, getChainsList, printChains } from '../../utils/utils.config'
 import { setMainnetStage } from '../utils'
 
 const contract = { contractName: 'OFTWrapper' }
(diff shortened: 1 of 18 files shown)


packages/stg-evm-v2/devtools/config/utils/utils.config.ts
```diff
@@ -257,9 +257,19 @@ function _filterChainsWithDeployments(chains: Chain[]): Chain[] {
...
+ * list (`NEW_CHAIN=foo-mainnet,bar-mainnet`). Returns an empty array when the
+ * var is unset or blank.
+ */
+export function getNewChains(): string[] {
+    return (
+        process.env.NEW_CHAIN?.split(',')
+            .map((c) => c.trim())
+            .filter((c) => c.length > 0) ?? []
+    )
 }
```
clauBv23, line 273: lets add a function to get the chainList directly and use it in all configs instead of duplicating 
```
const newChains = getNewChains()
    const chainsList =
        newChains.length > 0 ? newChains : process.env.CHAINS_LIST ? process.env.CHAINS_LIST.split(',') : []
```
```suggestion
}

export function getChainsList(): string[] {
    const newChains = getNewChains()
    return newChains.length > 0 ? newChains : process.env.CHAINS_LIST ? process.env.CHAINS_LIST.split(',') : []
}
```

packages/stg-evm-v2/devtools/config/mainnet/01/oft-wrapper.config.ts
```diff
@@ -12,8 +12,9 @@ export default async (): Promise<OmniGraphHardhat<OFTWrapperNodeConfig, unknown>
     // Set the stage to mainnet
     setMainnetStage()
 
-    const newChain = getNewChain()
-    const chainsList = newChain ? [newChain] : process.env.CHAINS_LIST ? process.env.CHAINS_LIST.split(',') : []
+    const newChains = getNewChains()
+    const chainsList =
+        newChains.length > 0 ? newChains : process.env.CHAINS_LIST ? process.env.CHAINS_LIST.split(',') : []
```
clauBv23: lets make a helper function to be used here 

https://github.com/stargate-protocol/stargate-v2/pull/538/changes#r3153822655

packages/stg-evm-v2/devtools/config/utils/asset.config.utils.ts
```diff
@@ -24,12 +24,15 @@ export default async function buildAssetDeploymentGraph(
...
-            // Chain doesn't support this token — return empty graph
+    // NEW_CHAIN mode: generate all connections to/from each new chain (and between
+    // new chains, without duplicating them). Node config is emitted for all chains;
+    // the framework only generates txs where on-chain state diverges from config.
+    const newChainNames = getNewChains()
+    if (newChainNames.length > 0) {
+        const newChainNameSet = new Set(newChainNames)
+        const newChains = supportedChains.filter((c) => newChainNameSet.has(c.name))
+        if (newChains.length === 0) {
+            // None of the requested new chains support this token — empty graph
```
clauBv23, line 39: I would also add a helper function to be used here and in messaging 

something like 

```
export function getNewSupportedChains(supportedChains: Chain[]): { newChainMode: boolean; newChains: Chain[] } {
    const newChainNames = getNewChains()
    const newChainMode = newChainNames.length > 0

    const nameSet = new Set(newChainNames) // trims + filters empty already
    if (nameSet.size === 0) return { newChainMode, newChains: [] }

    return { newChainMode, newChains: supportedChains.filter((c) => nameSet.has(c.name)) }
}
```

and then we only have to do 


```suggestion
    // NEW_CHAIN mode: generate all connections to/from each new chain (and between
    // new chains, without duplicating them). Node config is emitted for all chains;
    // the framework only generates txs where on-chain state diverges from config.
    const { newChainMode, newSupportedChains } = getNewSupportedChains(supportedChains)
    if (newChainMode {
        if (newSupportedChains === 0) {
            // None of the requested new chains support this token — empty graph
```

clauBv23:
fixed the comments in #544, let me know it you agree and feel free to adjust anything @ravinagill15 


skipped the new tests since they use rpcs, tested them locally, and all green

clauBv23 (approved):

PR #543, opened 2026-04-27
feat: update script to pin explicitly the default lib

clauBv23:


clauBv23:
closing in favor of #549

PR #545, opened 2026-04-30
validating default libs are expected (302) before explicitly setting …

ravinagill15:
…in config

### DIFF
--- a/packages/stg-evm-v2/devtools/config/utils/messaging.config.utils.ts
+++ b/packages/stg-evm-v2/devtools/config/utils/messaging.config.utils.ts
@@ -1,3 +1,4 @@
+import { EXPECTED_MESSAGE_LIB_VERSION } from '@stargatefinance/stg-definitions-v2'
 import {
     CreditMessagingEdgeConfig,
     CreditMessagingNodeConfig,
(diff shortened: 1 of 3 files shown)


packages/stg-evm-v2/devtools/config/utils/messaging.config.utils.ts
```diff
@@ -28,31 +29,69 @@ import {
...
+            // `defaultSendLibrary(dstEid)` / `defaultReceiveLibrary(srcEid)` need a peer eid.
+            // Querying a chain's default for itself isn't a real concept, so pick any other
+            // chain in scope. (LZ today uses the same lib across all destinations per chain.)
+            const peer = contracts.find((c) => c.eid !== contract.eid)
+            if (!peer) {
+                throw new Error(
+                    `Cannot fetch default message libraries for eid ${contract.eid}: at least one other chain must be in scope.`
+                )
+            }
+
```
ravinagill15, line 47: wondering if we should check defaults for all pathways instead of assuming they're uniform per chain. The Endpoint API stores defaults per destination, so it's technically possible for one chain's default to differ across destinations, and checking only one peer would silently miss that. That is not the case today, but could be in the future

ravinagill15: If we run in parallel using `Promise.all`, the fetches fan out across endpoints and version checks get de-duped to about 64 unique addresses, so performance should be more optimized

PR #552, opened 2026-05-13
Validation for pinned message libraries

ravinagill15:


Makefile
```diff
@@ -351,9 +353,12 @@ configure-mainnet:
...
 	$(CONFIGURE_OFT_WRAPPER) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/oft-wrapper.config.ts --signer deployer
 
-# 
+	# Validate that all messaging libs are pinned to the expected version on every chain
+	$(VALIDATE_PINNED_LIBS) $(VALIDATE_PINNED_LIBS_ARGS) --config ./hardhat.config.ts --stage mainnet
+
+#
 # This target will transfer the ownership of the mainnet contracts
-# 
+#
```
clauBv23, line 369: I think it should be a different script, note that after proposing the libs will not be correctly set yet

clauBv23 (approved):

Makefile
clauBv23: solved https://github.com/stargate-protocol/stargate-v2/pull/552/commits/cdd664e84a32c63e30d3741dc9f3dc6f5ed762c0

PR #562, opened 2026-06-04
Add unpeer to unwire logic

clauBv23:


Makefile
```diff
@@ -528,3 +534,11 @@ unwire-asset-mainnet:
 	# Remove assetId on TokenMessaging/CreditMessaging for the unwired chains
 	$(CONFIGURE_TOKEN_MESSAGING) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/token-messaging-asset.unwire.config.ts --signer deployer
 	$(CONFIGURE_CREDIT_MESSAGING) $(CONFIGURE_ARGS_COMMON) --oapp-config $(CONFIG_BASE_PATH)/credit-messaging-asset.unwire.config.ts --signer deployer
+
+unwire-chain-by-eid:
+	$(UNWIRE_MESSAGING_BY_EID) $(CONFIGURE_ARGS_COMMON) --dead-eids $(DEAD_EIDS)
+
+# Verify that deprecated chains are fully disconnected (hasPeer = false) from all active chains.
+# Reads deprecated EIDs and optional active chain list from chainsConfig/unwire/messaging.disconnected-check.yml
```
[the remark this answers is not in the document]
clauBv23: solved

packages/stg-evm-v2/devtools/tasks/checkMessagingDisconnected/checkMessagingDisconnected.ts
```diff
@@ -0,0 +1,186 @@
...
+            (r) => r.tokenMessaging !== 'disconnected' || r.creditMessaging !== 'disconnected'
+        )
+        const successes = results.filter(
+            (r) => r.tokenMessaging === 'disconnected' && r.creditMessaging === 'disconnected'
+        )
+
+        logger.info(`\n=== Disconnection Check Results ===`)
+        logger.info(GREEN(`✓ Fully disconnected: ${successes.length} / ${results.length}`))
+
+        if (failures.length > 0) {
```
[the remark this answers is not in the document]
clauBv23: fixed https://github.com/stargate-protocol/stargate-v2/pull/562/changes/f1037f98e192e4c98fb95b2d8dd20a0af20de53f

packages/stg-evm-v2/devtools/tasks/unwireMessaging/unwireMessaging.ts
```diff
@@ -0,0 +1,217 @@
...
+})
+
+/**
+ * Unpeer a fully-removed chain from all live messaging contracts by EID.
+ *
+ * Use when a chain has already been scrubbed from definitions (YAML + constant.ts deleted)
+ * so the normal graph-based unwire cannot include it. This task iterates all chains
+ * currently supporting messaging and calls setPeer(deadEid, bytes32(0)) on both
+ * TokenMessaging and CreditMessaging contracts where hasPeer(deadEid, null) = true
+ * (i.e. the peer is not yet zeroed out).
```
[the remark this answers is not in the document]
clauBv23: fixed https://github.com/stargate-protocol/stargate-v2/pull/562/changes/59999092a610786e3daf019ec820f235ed997846

packages/stg-evm-v2/devtools/tasks/checkMessagingDisconnected/messaging.disconnected-check.yml
```diff
@@ -0,0 +1,22 @@
+# Verifies that the listed deprecated chains are fully disconnected from all active chains.
+# Run with: STAGE=mainnet make check-messaging-disconnected
```
tinom9: I don't love the comment, this file is the config for the checker, not the checker, so it doesn't "verify".

packages/stg-evm-v2/devtools/config/mainnet/01/chainsConfig/unwire/messaging.disconnected-check.yml
```diff
@@ -0,0 +1,22 @@
+# Verifies that the listed deprecated chains are fully disconnected from all active chains.
+# Run with: STAGE=mainnet make check-messaging-disconnected
+#
+# deprecated_eids: EIDs of chains that should no longer be peered with anyone.
+#   Can be a single value or a list.
+#
+# active_chains (optional): names of chains to check against.
+#   If omitted, all chains currently supporting messaging are checked.
```
tinom9: I'd clarify that we're using EIDs for deprecated chains since we don't have `chainsConfig/*.yml` anymore.

As a side-note, I think it's cleaner to keep the source of truth in `chainConfig/`, that way it's less spread.

packages/stg-evm-v2/devtools/config/utils/unwire.config.utils.ts
```diff
@@ -105,12 +113,20 @@ export function loadMessagingUnwireConfig(): ResolvedMessagingUnwireConfig | und
...
 
+    const VALID_DIRECTIONS: UnwireDirection[] = ['from', 'to', 'both']
+
     const rules = rawConfig.rules.map((rule, index) => {
         if (!rule.chain || typeof rule.chain !== 'string') {
             throw new Error(`Messaging unwire rule missing 'chain' at ${configPath} (index ${index})`)
         }
         const allowedPeers = normalizeChainList(rule.allowed_peers, 'allowed_peers', configPath, 'Messaging unwire')
-        return { chain: rule.chain, allowedPeers }
+        const direction: UnwireDirection = rule.direction ?? 'both'
```
tinom9: I'd prefer throwing in the following error if unset or malformed rather than defaulting to `'both'`.

packages/stg-evm-v2/devtools/config/mainnet/01/chainsConfig/unwire/0-template-messaging.unwire.yml
tinom9, line 1: I believe we should be persisting these configurations for traceability?

packages/stg-evm-v2/devtools/tasks/unwireMessaging/unwireMessaging.ts
```diff
@@ -0,0 +1,273 @@
...
+                schema: TokenMessagingOmniGraphHardhatSchema,
+            })
+    )
+    subtask(
+        SUBTASK_LZ_OAPP_WIRE_CONFIGURE,
+        'Unwire token messaging: disable executor/DVN and remove peers',
+        (args: SubtaskConfigureTaskArgs<TokenMessagingOmniGraph, ITokenMessaging>, hre, runSuper) =>
+            runSuper({
+                ...args,
+                configurator: createConfigureMultiple(configureTokenMessaging, configureUnpeerEdges),
```
tinom9: Since `configureTokenMessaging` and `configureUnpeerEdges` are executed in the same run, can the following happen?

- Chain A is already unpeered.
- `configureTokenMessaging` tries to peer it.
- `configureUnpeerEdges` is a no-op, since it's already unpeered.

We get re-wirings instead of no-ops?

Should we drop `configureUnpeerEdges` and make `configureTokenMessaging` run `setPeer(0)`?

packages/stg-evm-v2/devtools/tasks/checkMessagingDisconnected/messaging.disconnected-check.yml
clauBv23: a018f1c24ae192824144b7ca75c47387eda2a793

packages/stg-evm-v2/devtools/config/utils/unwire.config.utils.ts
clauBv23: done e0eb91f99cea7e4a36e8d83805277bf6bfdec90b

packages/stg-evm-v2/devtools/config/mainnet/01/chainsConfig/unwire/messaging.disconnected-check.yml
clauBv23: done 22f78b51670f7e0f3a2beaa996e7012dc1e7865c

packages/stg-evm-v2/devtools/config/mainnet/01/chainsConfig/unwire/0-template-messaging.unwire.yml
clauBv23: we have not kept traceability for now but it can be done for now on

packages/stg-evm-v2/devtools/tasks/unwireMessaging/unwireMessaging.ts
clauBv23: solved 2ec20c67e0f9e8547ebd14bfd30a14c4cf3de9da

added a peerless messaging configurator that still applies the OApp executor/DVN config, but skips seiPeers

packages/stg-evm-v2/devtools/config/mainnet/01/chainsConfig/unwire/0-template-messaging.unwire.yml
tinom9: The main point is that we should be able to answer: "what chains were once in the mesh and have been unwired?"

`messaging.disconnected-check.yml` file could be the source of truth, but I don't think it's sufficiently explicit now.

What do you think?

clauBv23: agree, there are a couple of ways to check this, like the changelog or old PRs but it's not the same as having an explicit source of truth.

I updated `messaging.disconnected-check.yml` to include the EID, chain name, and a note per entry. The notes for older shutdowns might not be fully accurate since I don't remember all the details, but we can make them more meaningful going forward.

imo the best part of keeping this file is that we can check at any time that the mesh is properly unwired and there are no lingering dependencies, without having to dig through history.

clauBv23: There’s another detail to consider here: some chains have deprecated certain assets while the chain itself remains active.

For example, Base used to support EURC, but it was later removed from the list of supported assets, while other assets on the chain remain active.

How do you think we should reflect that? I don’t see it as a messaging disconnect, but I do think it’s worth gathering the information and deciding how best to represent it.

tinom9: Maintaining the actual config structure for wired chains and adding either chain-level or asset-level unwired flags sounds like the most straight-forward to me: you snapshot the config at the time of unwiring, and you treat both assets and chains (or whatever config you want to preserve) as append-only lists.

clauBv23: Maybe not the ideal solution, but it seems like the easiest for now.

I would keep `messaging.disconnected-check.yml` for fully unwired chains ( chains where the OApps are unwired), and handle unwired assets directly in the corresponding chain YAML under an `unwired-tokens` section.

Longer term, it would be better to consolidate these into a single structure, but I think we'd first need to define a model that properly covers all the different scenarios. 

Considering there are a lot of unknowns in the repo today, I’d prefer not to move in that direction yet and instead keep the implementation as straightforward as possible for now.

vanderspuy (approved):

tinom9 (approved):

PR #565, opened 2026-06-09
feat: allow defining the max msg size limit per path

clauBv23:

PR #572, opened 2026-06-25
Bump Polygon bus gas limit

clauBv23:
Summary:
- Bump Polygon TokenMessaging bus gas limit to 65k
- Add a patch changeset for stg-definitions-v2

Rationale:
- This is needed because Polygon [PIP-88](https://github.com/0xPolygon/Polygon-Improvement-Proposals/blob/main/PIPs/PIP-88.md) increased Polygon gas pricing, and Stargate bus executions started reverting from running out of gas under the previous 38k bus receive gas limit.
- The 65k value is based on the observed failing Polygon bus transaction path: it gives margin over the old 38k limit while keeping the change scoped to Stargate TokenMessaging bus `lzReceive` enforced options (`msgType = MSG_TYPE_BUS`).
- This does not change taxi messaging or LayerZero protocol-wide OApp defaults.


.changeset/polygon-bus-gas-limit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stargatefinance/stg-definitions-v2": patch
```
tinom9, line 2: #565 also added a changeset for `@stargatefinance/stg-evm-v2`, should we add here as well?

packages/stg-definitions-v2/src/constant.ts
```diff
@@ -3639,6 +3639,7 @@ export const NETWORKS: NetworksConfig = {
             },
             executor: EXECUTORS.LZ_LABS[EndpointId.POLYGON_V2_MAINNET],
             nativeDropAmount: parseEther('0.0324').toBigInt(),
+            busGasLimit: 65000n,
```
tinom9, line 3634: I'd leave a comment about _why_ we're bumping to 65k and based on what transaction we know 65k is a safe value.

Either in the code or in the PR description, either works.

.changeset/polygon-bus-gas-limit.md
clauBv23: the PR only changes definitions, so it is enough to just add a changeset for it.

packages/stg-definitions-v2/src/constant.ts
clauBv23: added to the PR description

tinom9 (approved):

PR #581, opened 2026-08-07
🛑 [shutdown] Glue mainnet 

clauBv23:
- Remove all chain references
- Deprecate both path directions


clauBv23:
Not adding changeset since it was done already in a [previous PR](https://github.com/stargate-protocol/stargate-v2/blob/736ef08db1d7639c43c4f9defe23c5d34bf8d449/.changeset/pink-singers-search.md)

tinom9 (approved):

